PR #107, opened 2015-01-03
Add initial version of qml/js plugin for kdevelop

joystein:
Builds, installs and seems to work, although not heavily tested.


mrueg:
Please also fix these repoman warnings:

 ebuild.minorsyn 2

dev-util/kdevelop-qmljs/kdevelop-qmljs-1.7.0.ebuild: Ebuild contains leading spaces on line: 27

dev-util/kdevelop-qmljs/kdevelop-qmljs-1.7.0.ebuild: Ebuild contains leading spaces on line: 28

metadata.missing 1

dev-util/kdevelop-qmljs/metadata.xml


For missing metadata, copy the skel.metadata.xml from the overlay's basedir and adjust if needed.

joystein:
I used RESTRICT="mirror" during testing to avoid having to wait so long. Will remove.

Is it really tab instead of spaces that is the recommended convention for indenting?

Will also add the metadata file.

joystein:
Corresponding bugzilla entry:

https://bugs.gentoo.org/show_bug.cgi?id=534464

PR #109, opened 2015-01-04
New ebuilds: kde-apps/kget, kompare and app-editors/kile

a17r:
Also, libkomparediff2 master switched to frameworks.
I'm sure kile needs more work, it currently does not build because of some okular headers mishap.


stikonas:
Shouldn't kile depend on kde-apps/konsole?

PR #114, opened 2015-01-11
[app-editors/kile:5] Re-organise dependencies using CDEPEND

a17r:
- move packages to CDEPEND where DEPEND is due
 - avoid cmake in RDEPEND
 - drop non-existing okular[postscript] use-dep
 - fix konsole RDEPEND
 - block app-editors/kile:4

### DIFF
--- a/app-editors/kile/kile-5.9999.ebuild
+++ b/app-editors/kile/kile-5.9999.ebuild
@@ -4,6 +4,8 @@
 
 EAPI=5
 
+CMAKE_MIN_VERSION="3.0.2"
+
 KDE_HANDBOOK=true
 EGIT_BRANCH="frameworks"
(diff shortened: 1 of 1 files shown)


a17r:
There's an automagic dependency on okular with viewerinterface.h, to be found in src/CMakeLists.txt, for "Live Preview Support":
---
# we now look for Okular with "viewerinterface.h"
FIND_PACKAGE( Okular )
---

That's actually the reason why it currently doesn't build, there's a problem with okular header visibility (a quick hack is to change the path inside viewerinterface.h so that it finds okularcore_export.h).

johu:
Please squash the commits and force push to your branch.

johu:
Thank you, this is rebased onto kde overlay master.

PR #115, opened 2015-01-18
kde-apps/pykde5, lokalize: add initial kf5 ebuilds

a17r:
* [kde-apps/pykde5]
 * derived from pykde4 ebuild (and metadata), though I usually have no business with python
 * PyQt5[gui, widgets] flags are the minimum required to *build* successfully, however what about [dbus, declarative, script, sql, svg]?
 * is arm patch still required?
 * EDIT: dropped sys-devel/libtool from DEPEND
* [kde-apps/lokalize]
 * master switched to kf5


johu:
Thanks for your work. This is rebased onto kde overlay master. 7a36c07883683dd53e210b04726bcdac868d0f35 was needed to fix file collisions.

PR #116, opened 2015-01-18
kde-apps/libkdegames, konquest: add kf5 ebuilds 

a17r:
Many games have frameworks branches already, one problem seems to be broken desktop files (Exec), however konquest run from terminal works fine.


johu:
Thanks for your work. This is rebased with reworded commit messages onto kde overlay master.

PR #117, opened 2015-01-18
[kde-frameworks/kauth] Re-enable PDEPEND on kde-plasma/polkit-kde-agent.

eliasp:
See also: 97ee921


johu:

PR #118, opened 2015-01-19
Qt5Test configure patches for konsole and dolphin, small fixes for kget and kile

a17r:
I noticed that the following packages would currently fail to configure without dev-qt/qttest installed:
* kde-apps/konsole:5
* kde-apps/dolphin:5

Both required a patch to CMakeLists.txt. I can try to upstream it in the next few days if the patches look good, unless someone else is quicker.

Minor ebuild fixes for kget and kile.


johu:
In general the commit message should be shorter and starts with a capital char. Like an email subject.

johu:
Thanks for your work. This is rebased onto kde overlay master.

PR #119, opened 2015-01-19
[net-im/ktp-text-ui] remove unneeded dependency

karolherbst:


johu:
Thanks for your work, This is rebased onto kde overlay master.

PR #120, opened 2015-01-20
[kde-misc/openofficeorg-thumbnail] add patch for kf5 support

karolherbst:
this is done through a little Qt4 to Qt5 port patch. The Project seems to be dead for 5 years now and I don't know if the author is still around.


johu:
Is this package realy needed? I realy dont like adding dead upstream packages and downstream patching.

karolherbst:
I am sure I got all runtime deps, though some build time deps could have slipped through. The thing with this is, that it won't be needed if you use Qt5 based calligra, because it will ship better thumbnailers.

But I think it makes sense to have a more lightweight sollution for users who don't want to install calligra just for that.

We already discussed forking the project, which I could do, because I already did the port, but I don't think I will care much later when Qt5 calligra is here.

karolherbst:
but which dependency do you miss explicitly? dev-qt/qtcore should be handled through kde5 eclass.

johu:
Thanks. This rebased onto kde overlay master. Note that ~x86 KEYWORD was wrong, because required dependencies are not keyworded yet.

PR #121, opened 2015-01-20
[net-irc/konversation] should depend on qca[openssl]

karolherbst:
DH and CBC (ECB too, but ECB doesn't matter) support are only provided through the qca-ossl plugin, which are the only ones supported by konversation. It makes no sense to enable crypt support without having the qca-ossl plugin enabled at all.


johu:
Looks weird indeed! But its more correct ;)

karolherbst:
updated

johu:
Thanks. This is rebased with formatted commit message body (line wrapping) onto kde overlay master

PR #122, opened 2015-01-21
Various packages: Make tests optional

a17r:
Patches should be upstreamed I guess.


johu:
Thanks for your work. This is rebased onto kde overlay master. Please try to upstream the patches.

PR #123, opened 2015-01-21
net-libs/telepathy-logger-qt: use right versions, also fix :5 ebuild

karolherbst:
sadly there is currently a compile error on the qt5 branch, but it should be safe nethertheless.


karolherbst:
Updated according to comments.

johu:
Thanks. This rebased onto kde overlay master with updated commit message. I dropped flex and bison dep in another commit.

PR #124, opened 2015-01-21
kate/kwrite: Make tests optional

a17r:
minor: fix KDE_HANDBOOK


johu:

PR #129, opened 2015-01-28
[net-im/ktp-auth-handler] fix Gentoo bug 537642

karolherbst:
fixes gentoo bug: https://bugs.gentoo.org/show_bug.cgi?id=537642


johu:
Thanks but i will fix this directly in tree, later on.

karolherbst:
yeah, no problem

PR #130, opened 2015-01-29
kapptemplate: make tests optional; gwenview: same plus USE="kipi,raw"; add libkipi kf5 along the way

a17r:


johu:
I would prefer if you keep RDEPEND="${DEPEND}" and please remove the "." from use flag desc.

johu:
Thanks. Rebased with reworded 2 commit messages onto kde overlay master.

PR #132, opened 2015-01-30
[kde-apps/svgpart] Add kf5 ebuild

a17r:

### DIFF
--- a/kde-apps/svgpart/svgpart-5.9999.ebuild
+++ b/kde-apps/svgpart/svgpart-5.9999.ebuild
@@ -1,0 +1,23 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Header: $
+
+EAPI=5
+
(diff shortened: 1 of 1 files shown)


kde-apps/svgpart/svgpart-5.9999.ebuild
```diff
@@ -0,0 +1,23 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Header: $
+
+EAPI=5
+
+EGIT_BRANCH="frameworks"
+inherit kde5
+
+DESCRIPTION="Svgpart is a kpart for viewing SVGs"
```
johu: We should not repeat the pkg name in  description, maybe "Kpart for viewing SVGs" should be enough.

johu:
Thanks. "[kde-apps/svgpart] Add KF5 version" rebased onto kde overlay master. 

I miss answers on the recruiting app ;)

PR #134, opened 2015-02-02
[eclass] Fix breakage by KDE_APPS_MINIMAL

a17r:
commit 6c68003fda689d2240ea45df4614166655f6b8ae made live packages depend on
e.g. >=14.12 instead of 5.9999. This change helps out as long as there exist
packages with separate frameworks branch. Alternative would be to add 5.9999
version to add_kdeapps_dep calls in every affected ebuild.


karolherbst:
works for me, thanks

PR #135, opened 2015-02-03
Remove EGIT_BRANCH, since since master is now frameworks based

jfmcarreira:


kensington:
Thanks, merged with PN prefix to commit message.

PR #136, opened 2015-02-03
[net-libs/libkeople] Follow upstream, drop dev-qt/qtwidgets (9e388c34).

eliasp:


kensington:
It looks like KPeopleWidgets is still installed by the same package, which still requires qtwidgets.

eliasp:
You're right. Thanks for noticing!

PR #138, opened 2015-02-05
[kde-plasma/kwin] kwin depends on <libinput-0.8.0

karolherbst:
upstream discussion: https://bugs.kde.org/show_bug.cgi?id=342893

currently kwin-5.2.0.1 fails to build with >=libinput-0.8.0, also the 5.2 branch disables libinput support completly if >=0.8.0 is found, hence we should depen on <0.8.0. The 5.3 will most likely support >=0.8.0.


johu:
Thanks. Rebased as "[kde-plasma/kwin] DEPEND on <dev-libs/libinput-0.8.0, bug #538870" onto kde overlay master.

PR #140, opened 2015-02-06
kde-base/*:4 updates to make kde-applications-14.12:4 installable

a17r:
The recently added 4.11.16 and 4.14.5 ebuilds were edited in place together with the respective live ebuilds.

I dropped kde-misc/krusader from my PR after noticing the overlay live ebuild had it already fixed, albeit not using eclass functions.

As far as overlay ebuilds are concerned, this PR should now be complete. An other example would be app-emulation/winetricks from tree which has an optional kde-base/kdialog dependency, I did not check all those.

Edit: Rearranged my PR so that this one covers only overlay changes.

### DIFF
--- a/app-editors/kile/kile-9999.ebuild
+++ b/app-editors/kile/kile-9999.ebuild
@@ -22,10 +22,10 @@ DEPEND="
 	x11-misc/shared-mime-info
 "
 RDEPEND="${DEPEND}
-	$(add_kdebase_dep kdebase-data)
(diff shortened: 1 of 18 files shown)


johu:
Could you please check if repoman is complaining if the package choice || is on one line instead of 4. This would reduce spacing a lot. On some released version i didnt see the live versions touched. Is this intentional?

a17r:
More PRs you shall receive. ;) Hope to have addressed all things as intended.

a17r:
obsolete by pkgmove

PR #141, opened 2015-02-06
[kde-apps/kdialog] fix RDEPEND for 5.9999 ebuild

karolherbst:


johu:
Thanks. This is slightly changed integrated in kde overlay master.

PR #142, opened 2015-02-06
[kde-apps/kdiamond,klines] Frameworks merged to master

a17r:


johu:

PR #143, opened 2015-02-08
[kde-apps/kblocks] Frameworks merged to master; [kde-apps/libkdegames] better patch

a17r:


johu:
Thanks. This is rebased onto kde overlay master. I would love to review answers on recruiting.gentoo.org from you. ;)

PR #144, opened 2015-02-10
[in-tree] kde-base/ category updates to add kde-apps/ compatibility

a17r:
Import of 4.14.3 ebuilds from main tree that needed to be changed so they depend on either kde-base/ or kde-apps/ version of a package (no live ebuild in overlay except kde-base/smokekde).

Edit: Rearranged my PR so the initial import is split from the actual changes done to the ebuilds.


a17r:
obsolete by pkgmove

PR #145, opened 2015-02-10
[kde-base/*meta] Revbump to depend on kde-apps/ where available

a17r:
kde-base/*meta revbumps which add_kdeapps_dep instead of kdebase where available, except for kde-meta (no change) and kdebindings-meta/kdepim-meta (both not a single kde-apps/ version available). I also added a mask file for the new meta ebuilds.

I am now able to _emerge -vp meta_ which would give me a clean upgrade path to 14.12:4 without blockers, and I enabled all kinds of use flags. Several kde-base/*:4 packages remain where no 14.12:4 update was made and all appears to work fine.


a17r:
Split out changes from import and revbump to improve review-ability.

johu:
I dont understand why this is needed at all? We have meta packages in kde-apps.

a17r:
Personally I don't use meta ebuilds (except for kdebase-runtime-meta as a dependency of kdebase-startkde), so admittedly I haven't looked too close at the kde-apps ones. However, as far as I understood, those pull in kde-apps/*:4 and kde-apps/*:5 packages simultaneously, whereas my updated kde-base/*meta ebuilds make sure users of KDE4 can stay on it a little while longer while getting all those 14.12.2:4 updates that happened.

a17r:
Now, when we look at kde-apps/kde-apps-meta, the top dependency (kde-apps/kate) already is kf5. Another example is kde-apps/gwenview:5 from kde-apps/kdegraphics-meta.

So there currently is no way to use kde-apps/*meta ebuilds without pulling in kf5/plasma5, unless those dependencies would be made optional.

johu:
What will be the result: You have to maintain kde-apps meta packages and kde-base meta packages all the time because they are moving stuff like crazy in the next month. @kensington what do you think?

a17r:
I haven't thought about 15.04 yet, but I can think of two solutions, both using #140 and #144.

######Absolute minimum
Only use kde-base/kdebase-runtime-meta of this PR, because it is a dependency of kde-base/kdebase-startkde.
Users lose the all-in comfort of meta, but can stay on KDE4+updates with a little effort.

######KDEAPPS_MAX (or similar)
Set inside meta ebuilds; add mechanism to add_kdeapps_dep() in kde4-functions.eclass to pin down all kde-apps dependencies to either slot "4/14.12" or version "<=14.12.9999".
KDE4 users keep their meta ebuilds and get all updates, but future updates can not wreak havoc.

kensington:
I definitely see the appeal, but understand the concerns about maintenance burden. A compromise could be to use || ( kde-apps kde-base ) deps (like in the other RRs) - the automagic in add_kdeapps_dep ensures that kde-base:4 will only pull in kde-apps:4.

We really need to consider our long-term migration plan (ie. pkgmoving apps or not). I think it'll go a long way to helping us sort out interim transition measures.

johu:
I  would love to see that :4 is dropped from kde-apps completly  and just package :5 in kde-apps. As this is a upstream release mess.

a17r:
I see kde-apps-14.12 mostly as a KDE4 upgrade, since kf5 is barely visible. Where it is, the apps do not have feature parity yet (looking at kate:5 and gwenview:5).

@kensington add_kdebase_dep will work for that as long as no other oddities such as kde-apps/kcontrol-14.12.2:**4** come along with 15.04, which has kde-plasma/khotkeys:**5** in RDEPEND. That's why I did not include it in kdebase-runtime-meta-4.14.3-r1 and it made me think of a more strict version requirement.

kensington:
@johu Unfortunately I guess that a full KF-5 only kde-apps will not be around for quite some time.

@a17r I'm a bit confused sorry - why would kcontrol:4 end up pulling khotkeys:5?

johu:
@kensington thats why i want to distungish between kde-apps and kde-base the other way arround kde-base as kdelibs4 based and kde-apps as frameworks based. So it would be much clearer. We could package the 14.xx 15.xx in both categories.

a17r:
@kensington That's how the dependency is set inside the ebuild. Unless it is just an || () away from kde-base/kcontrol, I didn't try that.

PR #146, opened 2015-02-10
[dev-util/kdevelop] Version bump

a17r:


johu:
Thanks, but bumped it myself in tree.

PR #147, opened 2015-02-12
Upstream repo name change (libkpeople → kpeople)

eliasp:


kensington:
Do we know if it's going to be shipped as kpeople too? Maybe we should rename.

johu:
Live ebuilds moved in c70969ad80cab48be5c5369c26f3605e5400ad81.

PR #148, opened 2015-02-12
[net-libs/libkpeople] repository got renamed

karolherbst:


karolherbst:
sorry, didn't see #147 but the other live ebuild has to be updated too

johu:
Live ebuilds moved in c70969ad80cab48be5c5369c26f3605e5400ad81.

PR #149, opened 2015-02-12
[app-cdr/k3b] fix failing patch when USE=-handbook

karolherbst:


a17r:
Strange, it works for me...

karolherbst:
I think it is a doc issue. the add_subdirectory( doc ) line is getting commented out and this triggers the epatch failure then.

karolherbst:
yeah, I tested it.
bug with USE=-handbook
no bug with USE=handbook

johu:

PR #150, opened 2015-02-12
Pkg move kde-apps/kxmlrpcclient -> kde-frameworks/kxmlrpcclient

a17r:
...and add the new dependency to kde-plasma/plasma-workspace as well.


johu:
Thanks

PR #152, opened 2015-02-15
[kde-plasma/plasma-workspace] Add dev-qt/qtconcurrent (upstream a462d12)

eliasp:

### DIFF
--- a/kde-plasma/plasma-workspace/plasma-workspace-9999.ebuild
+++ b/kde-plasma/plasma-workspace/plasma-workspace-9999.ebuild
@@ -57,6 +57,7 @@ COMMON_DEPEND="
 	$(add_frameworks_dep plasma)
 	$(add_frameworks_dep solid)
 	dev-libs/wayland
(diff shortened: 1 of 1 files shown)


johu:
I can not find any linking against qtconcurrent. Upstream issue of not needed dependency?

eliasp:
Sorry, had a typo in the commit ID in the title.
The [commit](http://quickgit.kde.org/?p=plasma-workspace.git&a=commitdiff&h=a462d12ab9bfdb11f6b3c418f2abdd6f06f773f5) which should've introduced this dependency.

QtConcurrent is listed in the link libraries and the code also makes use of it - no idea why the built code doesn't seem to link against QtConcurrent.

kensington:
I've seen a lot of similar QtConcurrent usage across various KDE stuff that never ends up in NEEDED. I never checked, but I guess it ends up getting inlined/optimised.

johu:
Thanks. Rebased with reworded commit message onto kde overlay master.

PR #155, opened 2015-02-16
[sys-block/partitionmanager] 1.1.1 Version bump

a17r:
Added KDE_LINGUAS="fi"
Dropped ENABLE_UDISKS2
New HOMEPAGE seems more appropriate (sourceforge hopelessly outdated)


sys-block/partitionmanager/partitionmanager-1.1.1.ebuild
```diff
@@ -0,0 +1,29 @@
...
+
+LICENSE="GPL-2"
+KEYWORDS="~amd64 ~x86"
+SLOT="0"
+IUSE="debug"
+
+RDEPEND="
+	dev-libs/libatasmart
+	>=sys-block/parted-3
+	sys-apps/util-linux
```
johu: Deps need sorting.

kensington:
Thanks, bumped straight in CVS. I wrote out $PN in $HOMEPAGE by hand, since technically we're not supposed to use variables there.

PR #157, opened 2015-02-18
[kde-apps/dolphin] Tests optional patch upstreamed

a17r:
Block by eclass, s/KDE/Plasma/


johu:

PR #160, opened 2015-02-19
[kde-apps/kcontrol] Should also work with kde-base/khotkeys

a17r:


johu:
Thanks. This is rebased onto kde overlay master with changed commit message.

PR #163, opened 2015-02-23
net-im/ktp-call-ui: added KF5 based version

karolherbst:
only compile tested


johu:
Thanks. Rebased onto kde overlay master with changed commit message ([net-im/ktp-call-ui] Add KF5 version).

PR #164, opened 2015-02-23
[kde-apps/okular] Fix tests-optional.patch

a17r:
Unfortunately though, upstream is negative about inclusion...


johu:
The patch applied is fine for gentoo, no need to update it downstream. It is sufficient when a enhanced version is applied upstream.

a17r:
It failed yesterday due to a CMakeLists chance, that's why I updated it.

johu:

PR #165, opened 2015-02-26
[RFC] kde-accounts-kcm: add RDEPEND for various services

karolherbst:
I am not sure if this is the best way to handle this. Ideally these flags should be handled inside net-im/telepathy-connection-managers or net-voip/telepathy-haze so that we can simply use? depend o them.


johu:
Yeah please file a bug or merge request via gentoo-rsync-mirror to net-im/telepathy-connection-managers this should not be handled in kde-accounts-kcm

PR #168, opened 2015-03-02
[kde-apps/dolphin] Switch branch to master

eliasp:
Now that Dolphin was split-off into its own repo, `master` is where the action happens.


johu:
Fixed with b1e2edb7265f0c057bda6a6da41334be78238427

PR #170, opened 2015-03-05
KTp: simple cleanups

karolherbst:


johu:
Thanks. This is rebased onto kde overlay master with slightly changed commit messages.

PR #171, opened 2015-03-05
[media-libs/qt-gstreamer] add multibuild support to 1.2.0 release + revbump

karolherbst:


johu:
Thanks. Revisoon bump is directly added to the tree. Changes to the live ebuild rebased onto kde overlay master.

+
+  05 Mar 2015; Johannes Huber <johu@gentoo.org> +qt-gstreamer-1.2.0-r1.ebuild:
+  Revision bump adds multibuild support by Karol Herbst <git@karolherbst.de>.
+

PR #173, opened 2015-03-06
KTp: fix a missing kpeople dep

karolherbst:


karolherbst:
removed the one commit

johu:
Thanks. This is rebased onto kde overlay master with changed commit message "[kde-apps/ktp-contact-list] DEPEND on kde-frameworks/kpeople"

PR #176, opened 2015-03-07
Alternative kde-applications packaging - :4 in kde-base, :5 in kde-apps

a17r:
As discussed, this has a few advantages over the existing all-in kde-apps/ approach:

- No ugly || () dependencies in reverse-deps required, best for non-KDE apps with KDE deps
- A clear separation of :4 and :5 apps, not just through slots (less error prone)
- KDE SC 4 users continue to work with kde-base/*meta ebuilds (no in-tree modifications required)
- kde-base/:4 applications could be moved to tree/testing already regardless of plasma-5 mask
- plasma-5 users can install kde-applications as intended by upstream through kde-apps/*meta
- A pkgmove could still be done at a later point, when plasma-5 is established in tree


johu:
@kensington lgtm, do we need to discuss this on next meeting first? http://wiki.gentoo.org/wiki/Project:KDE/Meeting/2015-03#KDE_Applications_5_transition_.2815_minutes.29

kensington:
@johu Looks fine but I think it's worth discussing anyway. I'll try to update the wiki with possible options and pros/cons in time for the meeting.

a17r:
A few things to add:
* In terms of effort, this is imo the best compromise since :4 reverse deps just won't break
* I don't like this solution for the load of ambigous ebuild names, but that's just an inconvenience
* kde-base/:4 ebuilds from applications could be moved to tree/testing already regardless of plasma-5 mask

johu:
@a17r i think we should just wait for the october release, i guess the kf5 port will be finished and we have clean transition...

a17r:
True, kde pim has probably received the most development of them all, and that's still 4.14. However, at least okular and umbrello have received feature upgrades, others many bug fixes - most notably baseapps, which at 4.14 aren't quite in a state where I would do without updates.

mrueg:
@a17r how would you handle applications that get migrated to kf5?
Assume we have kde-base/foo-15.04:4 to the tree and foo-15.08:5 is released. This way we won't be able to purge old versions, except we let people fall back to kde-base/foo-4.14.3:4, which is probably something we want to avoid.

a17r:
@mrueg First of all it needs to be clear if a 'seamless' upgrade to kf5 apps is even desired. If that is made possible, a lot of KDE SC 4 users could end up with questions how to make their kf5 apps look integrated without an available KCM in their systemsettings (:4 and :5 are not co-installable) or face conflicting dependencies of reverse-deps (we know people have all kinds of entries in package.keywords). The same is not possible with plasma-5, I guess at some point there is going to be a news item pointing to the upgrade guide, why not do the same when applications-4 are going to be removed from kde-base/?

I don't think the way kde-applications are packaged affects your ability to purge old versions. I reckon at least kde-base/foo-14.12:4 could eventually be stabilised - the small bugfixes diff makes it safe to replace respective 4.14.3 versions (purge those), and only few of those would remain since 14.12:4 is still an almost complete KDE4-apps release. 15.04 can be reviewed independently again, since it is going to be more of a mixed bag: Will it be worth packaging its :4 part at all (see @johu's suggestion on the meeting page) and eventually purge 14.12:4 where there is a 15.04:4 upgrade, or could its :5 part be made more complete by a snapshot here and there (thinking of dolphin and baseapps). Then again review, say, 15.08 when it is time. Completely purging :4 applications in favor of their kf5-ports won't be easy as long as there are reverse-deps that drag on their migration (think krusader, amarok), so kde-base/ and kde-apps/ will need to live side-by-side for some time anyway.

a17r:
Now that 15.04 was released, I'd like to take another look at this, since we've learnt a thing or two since last time:

Wrt...
* _Packaging 1:_ 15.04:4 is still a big enough chunk for me to care about; might get away with packaging only :5 in 15.08 though (KDE PIM planned!)
* _Packaging 2:_ If you want to provide a complete applications-15.04 as intended by upstream, you still need parts of 14.12:4 - things do get easier though if you don't strictly follow upstream (e.g. I can live with having a few kde-games:5 less until the next release - there are more pressing limitations)
* _Conflicts of interest:_ Changed for the worse, dolphin is still :4 so it continues to lack terminal integration, now dolphin-plugins:4 is left standing without newly ported kompare; dolphin:5 looks good enough for a snapshot but dolphin-plugins:5 imo not yet there; then there's gwenview:5 still without libkipi...
* _Purging:_ As a consequence, purging is not a simple case of dropping all 14.12:4, regardless of it being in kde-apps/ or kde-base/ - except for 14.12:5, which can all be dropped as soon as 15.04:5 is complete
* _Tree maintenance:_ A complete applications-xx.yy stable is far away, and at least ark-4.14.3, kdebase-data-4.14.3 and kstars-4.14.3 have already received backports, the list could grow longer until October; however a partial applications-xx.yy:4 might be just around the corner

KDE SC 4.14.3/4.11.18/4.14.7/14.12.3/15.04.0 is fun to write and still works best ;)

a17r:
obsolete by pkgmove

PR #177, opened 2015-03-08
KDE PIM 4.14.6 version bump

a17r:
...based on 4.14.5 ebuilds


johu:
Thanks. This is rebased as one commit onto kde overlay master.

johu:
Thanks. This is rebased as one commit onto kde overlay master.

PR #179, opened 2015-03-10
[kde-apps/dolphin-plugins] Add kf5 version

a17r:


johu:

PR #180, opened 2015-03-11
[kde-apps/kate,kwrite] Require kde-frameworks/kcrash

a17r:


johu:
Please follow Gentoo commit message wording, like "DEPEND on x" or add missing dependency x

a17r:
done

johu:

PR #181, opened 2015-03-12
[kde-apps/ktp-common-internals] Fix tests-optional patch

a17r:


karolherbst:
Thanks, I think I will include this in my ktp release branch if thats okay for you. I think I might also rework this in a way where no patch is needed at all anymore (if that's possible)

karolherbst:
as it seems, the patch can be somplified a lot and tests disabled unconditionally, because they are not integrated into ctest and require the X server, so yeah.

karolherbst:
The tests are useless anyway, that's why I decided to come up with something like that: https://github.com/karolherbst/kde/commit/c37ae1d90696c34180405c596907e4e9d30abd90

a17r:
Of course, even better ;) I'll still try to make them optional upstream.

PR #182, opened 2015-03-13
add Ktp-15.03.90 release

karolherbst:
everything except ktp-call-ui gets an kde application 15.04 release


johu:
This should be done as kde-applications bump.

PR #187, opened 2015-03-15
[kde-base/kmail] Fix runtime crash, upstream bug #345071

a17r:
Fix a regression in 4.14.6


johu:

PR #188, opened 2015-03-16
Patch inside src_prepare after eclass change

a17r:
`comment_add_subdirectory tests` otherwise happens before and thus breaks tests-optional patches. Two more to get upstream.


johu:
How about to upstream those patches and remove them downstream instead as you have upstream write access.

a17r:
Working on it - let's hope it's the last time they need to be touched.

johu:
I just removed the patches. Eclass should already handle it.

PR #192, opened 2015-04-07
[eclass] Invert version check logic for ECM

a17r:
It's easier the other way round when kde-frameworks/ is the only thing in tree.

Resolves conflicts with kde-plasma/ and kde-apps/


a17r:
OK I'm somewhat stuck here. I traded file conflicts for repoman anger (because of net-libs/telepathy-logger-qt) that would be solvable with an either-or DEPEND in there, but I also don't feel like changing an ebuild for three or so days until the pkgmove happens. Unless I should?

a17r:
Less lines & works better

a17r:
obsolete after pkgmove

PR #194, opened 2015-04-12
[kde-plasma/plasma-sdk] Add blocker on dev-util/plasmate:5

eliasp:


mrueg:
"+plasma-sdk - you may be familiar with plasmate which this includes
although it's not currently compiled"

Does it really block each other then?

eliasp:
The [plasmate repository](http://quickgit.kde.org/?p=plasmate.git) is gone and when I tried to build plasma-sdk, it collided with several plasmate related files.

So we should probably also remove plasmate completely.

kensington:
dev-util/plasmate removed in e6a8657e7eb140c6c689c0ba31cf9b1272205928.

kensington:
Thanks, merged in 898a76c93319c6aab647a8500db732e1c2565530 also applying to all versions and dropping the SLOT (collides with :4 too locally).

PR #196, opened 2015-04-16
kdiagram and kdepimlibs KF5 live ebuilds

karolherbst:
both libraries are optional dependencies for calligra and should be added to the tree before adding calligra


karolherbst:
I think I have to update the deps for kdepimlibs. I will do this tomorrow.

karolherbst:
should be good now

mrueg:
Merged, thanks!

PR #197, opened 2015-04-17
[kde-apps/gwenview] Version bump 15.04.0.

rinaldus:
Added kde-apps/gwenview-15.04.0 ebuild. It installable, I tested it in my local overlay. But I'm not sure, how correctly I added dependencies for "kipi" and "raw" USE-flags. I used traditional method because your "add_kdeapps_dep" operator tried to pull package from slot 5, but slot 4 needed. I didn't know, how to set correct slot and used traditional method "package_name-${PV}". Probably it needs to be corrected, but my ebuild merges and installs properly.


mrueg:
Thanks for your contribution, but we prefer full bumps for releases of KDE applications.
Superseded by https://github.com/gentoo/kde/pull/217 which adds the set of missing bumps for the kf5-based packages.

PR #198, opened 2015-04-18
[kde-misc/baloo-kcmadv] Fix HOMEPAGE and EGIT_REPO_URI

a17r:
gitorious.org shutdown - fixes bug #544848


mrueg:
Btw. is there still any need for this package or can we drop it?

a17r:
If I may quote upstream: https://bugs.kde.org/show_bug.cgi?id=233471#c23

"We removed the advanced interface for Baloo for configuring the filtering via GUI. Might not have been the smartest decision. We're discussing how to add it back. Till then there is an advanced kcm package."

;) ;)

mrueg:
Merged, thanks!

PR #201, opened 2015-04-19
[kde-apps] Purge Applications 14.12.3 ebuilds, but keep final KDE4 releases

a17r:
76 applications remain at 14.12.3 that got no 15.04.0:4 upgrade.


mrueg:
Let's introduce 15.04.0 completely first, then remove 14.12.3 as the next step.

a17r:
Yeah, this is not urgent, I just had the list available already from the splitting PR.

PR #202, opened 2015-04-19
[kde-apps/kde-l10n] Drop klipper handbook, add 15.04.0

a17r:
- Fix file collision with plasma-workspace-5.3
- 15.04.0 version bump done without actual system to test with currently, but looking at directory diffs seems fine


mrueg:
Merged, thanks!

PR #203, opened 2015-04-20
[net-irc/konversation] 1.6 version bump

a17r:


mrueg:
Please also drop "1.6" from your commit msg

a17r:
configure checks for Qca-qt5, not openssl - to be sure I also rebuilt it with qca:2[-openssl] successfully

kensington:
Indeed - most of the qca USE flags are for runtime plugins (replacing the old eg. app-crypt/qca-ossl).

mrueg:
Merged, thanks!

PR #215, opened 2015-05-02
skrooge, kmymoney version bumps

a17r:
- kmymoney-4.7.2
- skrooge-1.12.0


kensington:
Thanks, live ebuild changes merged and release versions pushed straight to main tree.

PR #217, opened 2015-05-04
[kde-apps] Add kf5-based part of KDE Applications 15.04

a17r:
Test build of kde-apps/kde-apps-meta-15.04.0 was all successful.


kensington:
Thanks!

PR #218, opened 2015-05-06
[kde-apps] Add missing 15.04.0 ebuilds

a17r:
TODO:
- Add plasma-telepathy-meta to kdenetwork-meta? When we do the same for 9999 however that would add a permanently defunct ktp-call-ui package build, unless we remove it from there as well.
- kde-base/kdelibs actually does not allow to use kde-apps/kde-env right now, so that should get a ||() dep


kensington:
I think we can remove kde-apps/kde-env completely - it's a custom Gentoo package and it's only pulled by kdelibs.

kensington:
It makes sense to add plasma-telepathy-meta to kdenetwork-meta since ktp is indeed part of kdenetwork upstream. I don't know about the ktp-call-ui issue though so can't comment there.

mrueg:
Merged, thanks!

PR #221, opened 2015-05-09
[kde-apps/kde-l10n] Split out KDE4-based l10n into separate ebuild

a17r:
Splits up the existing kde-base/kde-l10n ebuild into a KDE4- and KF5-based ebuild each, KF5 DEPENDs were actually completely missing in the old version.
- eclass deps/blockers changed
- kde-base/kdepim-meta can be satisfied with >=kde-apps/kde4-l10n-15.04.1 alone
- last commit finally adds back 'legacy' messages from 4.14.3 if USE=-minimal, that are otherwise missing from 15.04.1 tarballs, into kde-apps/kde4-l10n-15.04.1 so that it's a legitimate upgrade from 4.14.3 (that might be controversial ;)).

Though the tarballs are neatly organised into "4" and "5" directories, there's one catch to the splitting business - it reveals duplicate content as file collisions.

So, in essence, what's currently one ebuild:
- kde-apps/kde-l10n-15.04.1:4 (containing and depending on both 4 and 5)

becomes
- kde-apps/kde-l10n-15.04.1:5
- kde-apps/kde4-l10n-15.04.1:4

to stay compatible with current kde-apps packaging and avoid a name conflict with a future pkg move right away (instead of kde-base/kde-l10n-15.04.1).


mrueg:
Why not kde-l10n:4 and kde-l10n:5?

a17r:
What do we do with the version number? It's 15.04.1 for both.

mrueg:
a possible solution would be 15.04.1-r400 and 15.04.1-r500. but we also have oxygen-gtk{,3}.

a17r:
Yeah, that would work! Initially I thought about keeping the version at 4.14.3 and do revbumps, only having :5 progress with 15.04.x - but your idea would be better to maintain. What about maintenance anyway, would it be compatible with overlay scripts or extra steps required for the rev? I'm sure my current solution is fine in that regard.

kensington:
Thanks a lot!

PR #226, opened 2015-05-14
Add KDE Applications 15.04.1 release with KDE4 bits

a17r:
Repoman is happy, but some parts haven't been built yet (PIM could be nasty)


kensington:
Thanks!

PR #230, opened 2015-05-16
[kde-apps] Add kf5 snapshots

a17r:
This is a suggestion on how adding a few snapshots could improve the kf5 side of things (dolphin:5 with konsolepart:5; raw and kipi support in gwenview etc.; also thumbnails for dolphin:5). It is compatible with meta ebuilds and only needs a small change in kdegraphics-meta.


johu:
Commit hashes are not the correct way to do snapshots. Lets just wait for the the upstream release. :4 version are working in :5 too.

a17r:
Yeah, it wasn't the most serious of PRs, and personally I'm fine with using 9999:5 for those anyway. ;)

PR #231, opened 2015-05-16
[kde-frameworks/kwallet] Add USE=gpg and missing DEPENDs

a17r:
kde-frameworks/kwallet-5.10.0 needs a similar backport because it does the same find_package calls and links to it even though there is no gpgmepp-5.10.0 yet.

Package-Manager: portage-2.2.19


mrueg:
Merged, thanks!

PR #234, opened 2015-05-17
[kde-apps/kmix] Fix build with USE=-alsa

a17r:
As discovered and tested by ni1s

Package-Manager: portage-2.2.19


kensington:
Thanks!

PR #238, opened 2015-05-19
added Systemd kcm and latest version of kcmsystemd

karolherbst:
the github repository of kcmsystemd seems to be outdated now and moved to kde git with the new name systemd-kcm


mrueg:
So we probably should package move kde-misc/kcmsystemd then?

karolherbst:
I am not sure, but I guess so.

kcmsystemd master is KF5 based anyway now (the github repository)

karolherbst:
there are also some tags, I think I will create some more ebuilds before and do the pkgmove

karolherbst:
sadly the github repository wasn't imported into kde git, so that the tags aren't there and we still have to use github for the last stable release, which is 1.1.0 and KF5 based :/

johu:
Thanks. I have just bumped the version.

karolherbst:
@johu what about the live ebuild? It is most likely, that newer version will be called systemd-kcm, because the kde git repository is also named that way?

PR #243, opened 2015-05-25
[kde-apps/kde4-10n] Re-work path removal logic

a17r:
- Also fixes Bug 550346
- Tested with _all_ LINGUAS set for kde-apps/{kde,kde4}-l10n, no collisions left


kensington:
Really snazzy, thanks!

PR #250, opened 2015-06-05
[sets] Fix incorrect category.

Coacher:
Plus a followup update for the `kde-base/kde-l10n` to `kde-apps/kde4-l10n` package move.

Please also regenerate files under `Documentation/`.


kensington:
Thanks, manually merged since github is having some conflicts with master.

PR #254, opened 2015-06-07
[kde-base] Add KDE PIM 4.14.9

a17r:
At the same time fix kde-base/kleopatra EPATCH_SOURCE


johu:
Thanks!

PR #255, opened 2015-06-07
[Documentation] Fix metapackages and regenerate

a17r:
- kde-base/kdebase-meta -> kde-apps/kdebase-meta
- Drop SLOT=4 from oxygen-icons, "5" can be installed just as well


johu:
Thanks!

PR #256, opened 2015-06-08
[kde-apps/umbrello] Frameworks merged to master

a17r:
Actually supports both KDE4 and KF5 in master; only build the latter.


johu:

PR #258, opened 2015-06-09
kcmsystemd move and live ebuild

karolherbst:
according to https://github.com/rthomsen/kcmsystemd/issues/22 the old released were reuploaded with the new project name (allthough the content is still the same, that's why there is a S= for 0.7.0 and 1.1.0)

And because its the new name obviously we should pkgmove now.


karolherbst:
side note: the 0.6.1 version needs to be deleted then from the gentoo main repository, because this version wasn't reuploaded, but I think that's okay

karolherbst:
@rthomsen if you are fine with that ;)

karolherbst:
@rthomsen also, what about the metadata.xml stuff I left it blank in here, but if you want I could readd your stuff there.

rthomsen:
@karolherbst The 0.6.1 version is superseded by 0.7.0, so I don't see any reason why you should provide that. Sorry, I'm not familiar with Gentoo's build system. You want me to provide data for the fields in metadata.xml?

karolherbst:
I was thinking about using the old one, but I wanted to ask you first, because yours are in there

karolherbst:
also, where should bugs be reported now? I didn*t really find much information about the future of systemd-kcm.

rthomsen:
Yeah, systemd-kcm is in the process of moving to KDE infrastructure. So bugs should now be reported at: https://bugs.kde.org/
And as website you can use: https://projects.kde.org/projects/playground/sysadmin/systemd-kcm

karolherbst:
okay, in that case the stuff in the metadata.xml aren't relevant anymore, because bugs should direct go to kde.

johu:
Please squash the commits.

karolherbst:
@johu is it better now?

johu:

PR #260, opened 2015-06-09
[kde-apps/marble] Frameworks merged to master

a17r:
- Not sure about USE=kde yet (plasma instead?)
- USE="zip" currently does not work (requires quazip master which at this point does not build)


johu:

PR #264, opened 2015-06-11
[sys-block/kpmcore] New package

a17r:
libs split out of partitionmanager

@stikonas currently there seems to be a file collision between the two, I removed the file in partitionmanager for now


johu:

stikonas:
I have stopped installing that file in kpmcore (and I'm working on removing it completely there but it's not done yet). It's better to keep it in partitionmanager.

PR #267, opened 2015-06-13
[Documentation] Added 2 more entries

DogHackit:
Added 2 more entries that were necessary to prevent pykde4 from emerging


johu:
Fixed in 114dddcb725deb5c2a5a6f3ee12b9967086fb011

PR #274, opened 2015-06-21
[WIP] Add KDE PIM ebuild

a17r:
Monolithic ebuild as a base to split off all the parts.

### DIFF
--- a/kde-apps/kdepim/kdepim-9999.ebuild
+++ b/kde-apps/kdepim/kdepim-9999.ebuild
@@ -1,0 +1,152 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Header: $
+
+EAPI=5
+
+KDE_HANDBOOK=true
+KDE_TEST=true
+inherit kde5
+
+DESCRIPTION="Personal Information Management Suite"
+HOMEPAGE="https://www.kde.org/applications/office/kontact/"
+KEYWORDS=""
+
+PIM_FTS="akonadiconsole akregator blogilo console kaddressbook kalarm kleopatra kmail knotes kontact korganizer ktnef"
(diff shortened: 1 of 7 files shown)


kde-apps/kdepim/kdepim-9999.ebuild
```diff
@@ -0,0 +1,135 @@
...
+
+PATCHES=(
+	"${FILESDIR}/${PN}-revert-qwebengineview.patch"
+	"${FILESDIR}/${PN}-find-gpgme.patch"
+)
+
+src_prepare() {
+	kde5_src_prepare
+
+	cp "${FILESDIR}/FindGpgme.cmake" "${S}"/cmake/modules
```
johu: I would add FindGpgme.cmake to the find-gpgme.patch, it is easier to upstream the patch  then.

a17r: Maybe it should be added to ECM anyway, I've taken the file from kwalletd.

johu:
@a17r  Is there any progress on upstreaming the patches? To you talked with upstream about the repo split?

a17r:
The information I got was that gpgme should be found through gpgmepp, and upstream would be looking into it, but I still needed the patch a few days ago. Need to get back to upstream, but I should do that from my plasma5 system.

johu:
If we go with this solution we need a make.defaults with USE_EXPAND and the default enabled features.

a17r:
Pushed the latest state. Almost all the apps build independently of each other, except for kontact (which isn't of any use by itself anyway so I just forced kmail in that case) and kalarm which either needs the same treatment or get its kmail dependency made optional.

johu:
Thanks for your work. We will go with the monolithic approach for now, as we don't  need to re-introduce the eclass hacks for splitted version from kde:4 times.

PR #275, opened 2015-06-22
[kde-misc/krusader] Adjust DEPENDs

a17r:
Upstream commit 1f133f0be4868b9559632cf3a5669bef398f4334


johu:

PR #280, opened 2015-06-28
[eclass] Fixes

a17r:
for a happier repoman


a17r:
OK, failure is unrelated:

 selinux/package.mask:
  Mask systemd and deps in selinux profiles because no support in the policies.

johu:
@kensington what do you think?

kensington:
LGTM.

johu:

PR #282, opened 2015-06-29
Block kde4-l10n-4.14.3 for kde-plasma/ packages

a17r:
Block kde4-l10n-4.14.3 for kde-plasma/ packages, at the same time lift the mask on *l10n-15.04.3 to satisfy deps, this should resolve bugs like '#553572

To be merged after 15.04.3 release - unless I overlooked something?

Much better than constantly hacking at stable kde4-l10n-4.14.3 ebuild, might as well revert the [minimal] flag afterwards.

### DIFF
--- a/eclass/kde5.eclass
+++ b/eclass/kde5.eclass
@@ -136,8 +136,8 @@ case ${KDE_AUTODEPS} in
 
 		if [[ ${CATEGORY} = kde-plasma ]]; then
 			RDEPEND+="
+				!<kde-apps/kde4-l10n-15.04
 				!kde-apps/kde4-l10n[-minimal]
-				!kde-base/kde-l10n:4[-minimal(-)]
 			"
 		fi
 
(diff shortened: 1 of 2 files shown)


a17r:
Hmm that's true, I only looked at the overlay. Anyway, how far away is that, and wouldn't an exception to l10n be possible? ;)

There are several file collisions in tree right now that would be solved by kde-applications-15.04:4 (and a block on the affected 4.14.3 version instead of messing with minimal flags).

johu:
@kensington You wanted start moving apps to the tree. Please do it together with this pr!

kensington:
@johu unfortunately it is taking much longer than expected as most :5 packages have outdated deps.

johu:
@a17r Can we close this PR as bug #553572 is fixed by Plasma 5.3.2 itself?

a17r:
yes we can! good that it was fixed.

PR #283, opened 2015-06-29
[kde-plasma/breeze] USE kde4 -> qt4

a17r:
Setting affects the look of all Qt4 applications in a plasma session, so use of more common qt4 flag seems appropriate.


johu:
Thanks. :+1:  Rebased onto kde overlay master.

PR #286, opened 2015-07-04
media-gfx/synnefo update

a17r:
kde-misc/kolor-manager-1.0.2 does not work with oyranos-9999 anymore

### DIFF
--- a/media-gfx/synnefo/metadata.xml
+++ b/media-gfx/synnefo/metadata.xml
@@ -2,7 +2,10 @@
 <!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
 <pkgmetadata>
 	<herd>kde</herd>
+	<upstream>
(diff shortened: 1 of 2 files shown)


a17r:
I wanted to have something to actually make use of Qt4-based synnefo after kolor-manager-9999 ebuild switched to kf5; kolor-manager-1.0.2 is too old and also incompatible to oyranos master (prerequisite of synnefo right now).

johu:
I am realy against all those snapshots in an overlay when you can build live version. Snapshots are solution for the main tree when for example there is a need for to keeping a package alive or make it working again.

a17r:
Well, this is mostly for testing and required synnefo/oyranos deps are live as well anyway - it's just that 4.9999 was dropped already. Would you mind bringing back that one or shall I keep it private?

johu:
I would vote for keeping it private then, as we are trying to port the deps forward.

johu:
Thanks, this is rebased onto kde overlay master. Did i mention already "do your quizzez"?

PR #287, opened 2015-07-05
[kde-base/korganizer] Build handbook

a17r:
Fixes bug 550500


johu:
Is it possible to use one patch for all versions?

a17r:
No, the CMakeLists.txt was changed at some point in between, but korganizer-4.10.0-handbook.patch will most likely work for current stable 4.14.3 as well (which the bug is about)

johu:
Thanks. This is rebased ontp kde overlay master.

PR #289, opened 2015-07-06
[kde-base/korganizer] doc dir not extracted if USE=-handbook

a17r:
...always test the other way round...


johu:

PR #290, opened 2015-07-06
kdevelop (and related) updates, new package dev-util/kdevelop-sql

a17r:
- kdevelop-4.7.1-ninja-optional.patch successfully tested with kdevelop-4.7.1 in tree
- kdevelop-qmake-9999:5: add kf5 version

### DIFF
--- a/dev-util/kdevelop/files/kdevelop-ninja-optional.patch
+++ b/dev-util/kdevelop/files/kdevelop-ninja-optional.patch
@@ -1,0 +1,9 @@
+--- a/projectbuilders/CMakeLists.txt	2015-02-02 15:19:46.000000000 +0100
++++ b/projectbuilders/CMakeLists.txt	2015-07-06 22:51:22.083293634 +0200
+@@ -1,5 +1,5 @@
+ 
+ add_subdirectory(makebuilder)
+-add_subdirectory(ninjabuilder)
++ecm_optional_add_subdirectory(ninjabuilder)
+ ecm_optional_add_subdirectory(cmakebuilder)
+ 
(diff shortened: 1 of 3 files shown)


johu:
I would vote to keep it in SLOT 0 as the tree version is not moved. So how about to properly slotmove it there properly first?

johu:
I have applied the patch "[dev-util/kdevelop-qmake] Add kf5 version" and cleaned out some kdelibs4 kdev plugins. Could you please rebase your PR onto the latest changes in kde overlay master.

johu:
I have applied patches "[dev-util/kdevelop-php] Fix DEPENDs", "[dev-util/kdevelop-pg-qt] Slotmove from 0 to 5".

The ninja use flag is something for the tree as the kdelibs4 live version will be cleaned out very soon or can this also applied to :5? The -sql package is interesting when there is a frameworks port available.

a17r:
Actually looks like something similar can be done for kdevelop:5 but the patch needs to be changed a bit.

Will file PR for kdevelop:4 patch on gh mirror, do you mind me doing the same for kdevelop-sql:4 (release version)?

johu:
I think we could wait for a "real" release or at least we need to put the tarball on a mirror.

a17r:
So, I've ported the patch to kdevelop:5 but one can debate whether it is that useful after all. Startup didn't spam about missing ninja in that version, and it is reasonably small.

johu:
I have applied "[dev-util/kdevelop-css] New package" patch :+1:  and added the package to the yeserday created kdevelop-live set :)

johu:
Thanks patch applied.

PR #291, opened 2015-07-07
[kde-plasma/plasma-desktop] Split up USE=usb into evdev,logitech

a17r:
Makes virtual/libusb:0 optional through USE=logitech
Keeps kcm_input optional through USE=evdev (and re-adjusts DEPENDs)

### DIFF
--- a/kde-plasma/plasma-desktop/metadata.xml
+++ b/kde-plasma/plasma-desktop/metadata.xml
@@ -3,6 +3,7 @@
 <pkgmetadata>
 	<herd>kde</herd>
 	<use>
+		<flag name="evdev">Enable configuration module for desktop input devices.</flag>
 		<flag name="gtk2">Add support for gtk+2-based applications using legacy xembed systray icons via libappindicator.</flag>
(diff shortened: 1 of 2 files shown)


johu:
Please rebase latest changes. I have already removed the patch in kde overlay master.

a17r:
Yes, certainly makes a lot of sense to be enabled by default. Removed evdev USE flag again to avoid confusion, it can be added any time should there be some libinput alternative or use of plasma without mouse in the future. Also moved to DEPEND only as just a header is required.

Also dropped the disabling of packagekitqt5 - thought it could prevent trouble by the time such a package goes into portage.

a17r:
Another day another situation - upstream made libusb:0 dependency optional (it's for very old Logitech devices in fact) and now evdev flag does what was previously enabled by usb.

johu:
Lets wait for the review then.

johu:
Thanks. This is rebased onto kde overlay master. Quizzzzzzzzzzzzzzzzzz! :)

PR #295, opened 2015-07-10
[kde-apps/libkipi] Fix DEPENDs

a17r:


johu:
Thanks. Its weekend, a good opportunity to work on some questions.

PR #296, opened 2015-07-12
[media-gfx/kgraphviewer] Add kf5 version

a17r:

Currently doesn't seem to build with Qt 5.5, but upstream is informed already (it's an easy fix).


kensington:
Thanks!

PR #297, opened 2015-07-16
Fix PIM collisions, net-libs/libkgapi missing DEPEND

a17r:


johu:
Thanks and please do from time to time a little bit quiz work.

PR #298, opened 2015-07-16
[kde-plasma/kwin] Add missing dev-qt/qtwayland dep.

eliasp:

Otherwise, `kwin_wayland` will report on startup:
```
default unknown: This application failed to start because it could not find or load the Qt platform plugin "wayland".

Available platform plugins are: minimal, offscreen, xcb.

Reinstalling the application may fix this problem.
zsh: abort (core dumped)  kwin_wayland
```


johu:
Thanks. Patch applied.

PR #299, opened 2015-07-17
Fix kde-apps/kdesdk-meta dependencies

Vindex17:
* I removed kompare and libkomparediff2 because they are already pulled in as dependencies for dolphin-plugins, which selects the correct slot and version for both packages, so there's no need to have them here. This also fix this conflict (both on world update and kdesdk-meta installation):

> \* Error: The above package list contains packages which cannot be
> \* installed at the same time on the same system.
>
>  (kde-apps/kompare-14.12.3:4/14.12::kde, installed) pulled in by
>    \>=kde-apps/kompare-14.12.3:4[aqua=] (>=kde-apps/kompare-14.12.3:4[-aqua]) required by (kde-apps/dolphin-plugins-15.04.3:4/15.04::kde, installed)
>
>  (kde-apps/kompare-15.04.3:5/5::kde, ebuild scheduled for merge) pulled in by
    \>=kde-apps/kompare-14.12.3 required by (kde-apps/kdesdk-meta-15.04.3:5/5::kde, installed)
>
>  (kde-apps/libkomparediff2-15.04.3:5/5::kde, ebuild scheduled for merge) pulled in by
>   \>=kde-apps/libkomparediff2-14.12.3 required by (kde-apps/kdesdk-meta-15.04.3:5/5::kde, installed)
>    \>=kde-apps/libkomparediff2-15.04.3:5 required by (kde-apps/kompare-15.04.3:5/5::kde, ebuild scheduled for merge)

* Thanks to the first modification, there are no conflicts at all, so there's no need for the "relaxed_deps" useflag, which also was almost useless because it was impossible to have it disabled, again because of the above-mentioned conflict.

### DIFF
--- a/kde-apps/kdesdk-meta/kdesdk-meta-15.04.3.ebuild
+++ b/kde-apps/kdesdk-meta/kdesdk-meta-15.04.3.ebuild
@@ -9,7 +9,7 @@ inherit kde5-meta-pkg
 DESCRIPTION="KDE SDK - merge this to pull in all kdesdk-derived packages"
 HOMEPAGE="http://www.kde.org/applications/development"
 KEYWORDS="~amd64 ~x86"
-IUSE="cvs relaxed_deps"
+IUSE="cvs hard_deps"
 
 RDEPEND="
 	$(add_kdeapps_dep dolphin-plugins)
@@ -23,9 +23,6 @@ RDEPEND="
 	$(add_kdeapps_dep okteta)
 	$(add_kdeapps_dep poxml)
 	$(add_kdeapps_dep umbrello)
-	relaxed_deps? ( $(add_kdeapps_dep kompare '' '14.12.3')
-		$(add_kdeapps_dep libkomparediff2 '' '14.12.3') )
-	!relaxed_deps? ( $(add_kdeapps_dep kompare)
-		$(add_kdeapps_dep libkomparediff2) )
 	cvs? ( $(add_kdeapps_dep cervisia) )
+	hard_deps? ( $(add_kdeapps_dep kompare) )
 "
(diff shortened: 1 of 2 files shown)


johu:
I disagree here, meta packages should contain the whole set of packages to be part of.

kensington:
I agree with @johu. Hopefully dolphin will be KF5-based in 15.08 which should improve the situation.

Vindex17:
Then how about this?

johu:
Same comment as before.

Vindex17:
And now?
P.S. I remember you that the ebuild does not work the way it is now

a17r:
What about sth like:

|| (
$(add_kdeapps_dep {kompare,libkomparediff2})
$(add_kdeapps_dep {kompare,libkomparediff2} '' '14.12.3')
)

Vindex17:
That won't work either, because il be set again as '>=14.12.3', so it will try to install the 15.04.3 causing the block

a17r:
Then so:

|| (
$(add_kdeapps_dep {kompare,libkomparediff2})
>=kde-apps/{kompare,libkomparediff2}-14.12.3:4
)

Vindex17:
updated again to fix travis issue

Vindex17:
@a17r You last solution is almost the same as my last one, except it's quite redundant since
dolphin-plugins automatically sets *>=kde-apps/{kompare,libkomparediff2}-14.12.3:4* using the correct eclass

a17r:
yeah but it takes care of 'meta packages should contain the whole set of packages' and gets rid of the use flag

johu:
Use flag is removed within commit f8289a470d716fc5514729d058290132a738127a

PR #300, opened 2015-07-17
[kde-apps/kpeoplevcard] New package

a17r:


johu:
Thanks :+1: . Rebased onto kde overlay master. Please notice: its weekend, quiz time.

PR #301, opened 2015-07-18
[kde-apps/picmi] Master switched to kf5

a17r:


johu:
Thanks. Rebased. Hopefully you are doing some quiz work ;).

PR #302, opened 2015-07-18
[kde-plasma/plasma-sdk] Patch out bogus upstream dep

a17r:
Patch works the same for 5.3.2.


kensington:
Thanks!

PR #304, opened 2015-07-21
[kde-apps/dolphin-plugins] Fix branch, missing DEPENDs

a17r:
- dropped libkonq, required only for not yet ported hg
- fix Qt slot


kensington:
Thanks

PR #310, opened 2015-07-22
[kdevelop-cppcheck] New package

a17r:


johu:
Thanks. There is also a valgrind plugin available ;) ... and of course some questions to answer.

PR #314, opened 2015-07-24
[kde-apps] Add KDE Applications 15.08 stable branch

a17r:
Ebuilds copied from 9999 (only where 15.04 existed), then added some


kensington:
Thanks!

PR #315, opened 2015-07-24
[eclass] Avoid confusion over system configuration

a17r:
E.g. https://forums.gentoo.org/viewtopic-p-7786236.html#7786236

Since kde-apps/kdebase-runtime-meta is a dependency of kdebase-startkde, I figured the latter is redundant here anyway.


kensington:
Thanks!

PR #316, opened 2015-07-25
[kde-apps/kdebase-runtime-meta] Fix conflict with kf5-based apps

a17r:
See also https://github.com/gentoo/gentoo-portage-rsync-mirror/pull/183


johu:
Thanks. Any questions to review?

PR #319, opened 2015-07-26
[kde-apps/libkdegames] Fix patch for bug 555980

a17r:
my fault...


johu:
Thanks, rebased onto kde overlay master. Please upstream the patch ASAP as we dont want to ship this forever.

PR #320, opened 2015-07-26
[kde-apps] 15.08 - Some adjustments

a17r:


a17r:
Actually I think the libkonq dependency in ark is a bit backwards; I think it just enables the compress action within dolphin if ark is installed, which already has the libkonq dependency. Likewise, setting USE=dolphin for ark does nothing if dolphin is not installed. Maybe this would be best solved in a dolphin pkg_postinst message?

a17r:
OK my memory was wrong, libkonq needs to be at the same version as ark.

johu:
Thanks, rebased.

PR #321, opened 2015-07-27
[net-misc/smb4k] Add kf5 version

a17r:


johu:
Thanks. This is rebased onto kde overlay master. Please work on the quiz, it is the same fun as ebuild work.

PR #323, opened 2015-07-27
[kde-apps/kde-apps-meta] Fall back to released kde-l10n versions

a17r:
Fixes repoman complaint and USE=nls in live ebuilds


johu:

PR #324, opened 2015-07-27
[kde-base/kdepimlibs] Add USE=minimal to fix block caused by kde-apps/kwalletd

a17r:
At the same time this should go into plasma profile package.use.mask:

kde-base/kdepimlibs:4 minimal -handbook -prison

### DIFF
--- a/kde-apps/kdepimlibs/kdepimlibs-15.08.49.9999.ebuild
+++ b/kde-apps/kdepimlibs/kdepimlibs-15.08.49.9999.ebuild
@@ -61,7 +61,7 @@ DEPEND="${COMMON_DEPEND}
 	dev-libs/boost
 "
 RDEPEND="${COMMON_DEPEND}
-	!kde-base/kdepimlibs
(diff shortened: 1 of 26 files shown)


johu:
As i know both kwalletd can are co-installable, is this realy needed?

a17r:
kwalletd is co-installable, but the trouble starts when you set USE=gpg, which pulls in kde-base/kdepimlibs:4, and its akonadi stuff collides with kde-apps/kdepimlibs:5

johu:
:-1: On that huge change for disabling one use flag.

a17r:
Wouldn't that make existing gpg wallets from upgrading KDE4 users unusable?

a17r:
Possible alternative:

Introduce kde-apps/legacy-kdepimlibs:4 and make kde-apps/kwalletd:4 depend on it optionally, the blocker on kde-base/kdepimlibs:4 through :5 will settle the dependency and remove the need to touch all those old pim ebuilds.

a17r:
Closing as split package is much better imo.

PR #325, opened 2015-07-28
Add kde-apps/libkeduvocdocument to kdeedu-live

a17r:


johu:

PR #326, opened 2015-07-28
[kde-apps/gpgmepp] Split out gpgmepp from kdepimlibs:4

a17r:
See also https://github.com/gentoo/kde/pull/324

Serves as an alternative dependency for kde-apps/kwalletd:4 - so we can keep USE=gpg and do not run into conflict of kde-base/kdepimlibs vs. kde-apps/kdepimlibs


johu:
Thanks, rebased onto kde overlay master. Please keep up the good work and make some progress on the quiz.

PR #329, opened 2015-07-30
[kde-apps/kget] DEPEND on kde-apps/gpgmepp:4 alternatively

a17r:
one more use case


johu:

PR #332, opened 2015-07-31
[kde-apps] Remove conflicting KDE4-based RDEPENDs from meta

a17r:
kgpg - depends on kde-base/kdepimlibs:4
kopete - depends on kde-base/kdepimlibs:4
kuser - depends on kde-base/kdepimlibs:4


johu:

PR #333, opened 2015-07-31
[eclass] Fix KDE_SCM=svn

a17r:


johu:
Thanks. Enough ebuild work for you, please proceed with quizzez ;)

PR #336, opened 2015-08-01
media-gfx/digikam 4.12.0 version bump

a17r:


a17r:
one more change, added slot depend to libmediawiki in preparation of :5.

johu:
@akhuettel please review, as it's your maintained package.

a17r:
I'm not so sure about dropping SLOT in kde-apps/libkface - sure it's redundant when using kde4-base, but media-libs/libkface makes use of subslots so maybe some some release in the future would make it necessary in kde-apps/libkface as well (same for other libs).

johu:
But then we then we maybe can find a eclass solution?

a17r:
Yes, especially since there are a few candidates in digikam world for subslots. Also for depending ebuilds, some means to define a slot operator would be nice to have in _add_kdecategory_dep()

johu:
Thanks. Rebased onto kde overlay master. Please keep your quiz in mind.

PR #337, opened 2015-08-02
[media-plugins/kipi-plugins] Add kf5 version

a17r:
...including
[net-libs/libmediawiki] Add kf5 version
as an optional dependency.

I've kept the lines for release-version intact for reuse when the first tarball arrives.


johu:
Please rebase

johu:
Thanks.

PR #338, opened 2015-08-03
[dev-util] Drop kdelibs4support from DEPEND as did upstream

a17r:


johu:

PR #340, opened 2015-08-04
[kde-frameworks/baloo] Drop kdelibs4support from DEPEND

a17r:
...just in time


johu:
Sorry done this already before bumping 5.13.0.

PR #341, opened 2015-08-07
[kde-apps/kde-l10n] Always block older kde4-l10n

a17r:


johu:

PR #344, opened 2015-08-10
[kde-apps/kdepim] Add missing DEPEND

a17r:


johu:

PR #345, opened 2015-08-12
kde5.eclass: Fix release version handbook bug 557390

a17r:
Do not accidentally comment any directory if KDE_DOC_DIR not exists


kensington:
Thanks!

PR #347, opened 2015-08-14
kde-apps/kdewebdev-meta: Drop kde-apps/klinkstatus from RDEPEND

a17r:
It depends on conflicting kde-base/kdepimlibs:4


johu:
Thanks.

PR #348, opened 2015-08-16
kde-plasma/plasma-workspace: Backport consolekit2 support

a17r:


johu:
Thanks. Will nake a lot of openrc user happy. A lot of devs would be happy if would work on the quiz.

PR #349, opened 2015-08-17
kde-apps/kscreengenie: Add release version

a17r:
...and make use of it in kdegraphics-meta as an alternative to ksnapshot. kscreengenie is going to be renamed to ksnapshot for kde applications 15.12.


johu:

PR #351, opened 2015-08-19
kde-apps/kdeedu-meta: Relax RDEPEND on kde-apps/marble

a17r:
While marble was ported to kf5 in 15.08, libkgeomap:4 and packages depending on it continue to pull in kde-apps/marble:4. Let the meta ebuild go out of their way.


johu:
Thanks.

PR #358, opened 2015-08-25
kde-apps/kdebase-meta: Add USE=minimal to hide kde-apps/plasma-apps

a17r:
Though no blocker, it has no use on a plasma 5 system and needlessly pulls in kde-apps/libkonq:4. In plasma 5 it is part of plasma-desktop.

However, kde-apps/libkonq:4 still has too many other dependants to be dropped, so plasma-5 users who want better integration by libkonq:5 better simply not use kdebase-meta and kdenetwork-meta.

### DIFF
--- a/kde-apps/kdebase-meta/kdebase-meta-15.08.0.ebuild
+++ b/kde-apps/kdebase-meta/kdebase-meta-15.08.0.ebuild
@@ -8,7 +8,7 @@ inherit kde5-meta-pkg
 
 DESCRIPTION="Merge this to pull in all kdebase-derived packages"
 KEYWORDS="~amd64 ~x86"
-IUSE="+wallpapers"
+IUSE="minimal +wallpapers"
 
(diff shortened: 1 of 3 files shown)


kensington:
Speaking about packages that are only useful for KDE 4, I just realised kdebase-meta doesn't offer any way to install Plasma 4 anymore. So, I am wondering if we should revert 25da2233af7acd6e2a743d4d04dd0acbcec88413 and put this behind that.

a17r:
Hmm I wouldn't bring back the kde-workspace deps, rather 
- have them in a separate plasma-desktop:4 ebuild.
- unless in fact kdebase-meta was the only way to install Plasma 4, revert 25da2233af7acd6e2a743d4d04dd0acbcec88413 and pkgmove to plasma-meta, then depend on said plasma-desktop:4 if USE=-minimal?

Just thinking out loud ;)

PR #364, opened 2015-08-30
Remove KDELibs4support dependencies in kdepim ebuilds

MJDSys:
As I was upgrading my system, I got some build errors regarding these packages due to missing dependencies.  These two are the ones found so far.

### DIFF
--- a/kde-apps/kimap/kimap-15.08.0.ebuild
+++ b/kde-apps/kimap/kimap-15.08.0.ebuild
@@ -13,6 +13,10 @@ LICENSE="LGPL-2+"
 KEYWORDS="~amd64 ~x86"
 IUSE=""
 
+PATCHES=(
+	"${FILESDIR}"/kimap-15.08.0-remove-kdelibs4support.patch
+)
+
 RDEPEND="
 	$(add_frameworks_dep kcodecs)
 	$(add_frameworks_dep kcoreaddons)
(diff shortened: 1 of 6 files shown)


kde-apps/kimap/kimap-15.08.0.ebuild
```diff
@@ -18,6 +18,7 @@ RDEPEND="
 	$(add_frameworks_dep kcoreaddons)
 	$(add_frameworks_dep ki18n)
 	$(add_frameworks_dep kio)
+	$(add_frameworks_dep kdelibs4support)
```
a17r: Dependencies are sorted alphabetically.

+others

MJDSys:
@a17r Fixed

kensington:
We should probably patch these out.

In kde-apps/kimap it's only used for tests, and in kde-apps/kpimtextedit it's not used at all (and is properly removed upstream in master).

a17r:
kimap was taken care of upstream it seems: https://quickgit.kde.org/?p=kimap.git&a=blobdiff&h=c4af8c85d1511e298428f0f50a05fc492fec9487&hp=d2beff41afc1fb5ebcd369eda0f4d30beec7c6f1&hb=bf73c02636b55197e7c7ff870739c98cb200ef1a&f=CMakeLists.txt

MJDSys:
Ok, I took patches from both upstream that remove KDELibs4Support, and applied it to the 15.08* ebuilds.  The live ebuilds are untouched, since upstream has removed it.

kensington:
Thanks, rebased as 20994322e0407c6423f249debbe98a21e7d941fd and f94013ea73102002a0141e27ee3c032887ad88a8.

PR #372, opened 2015-09-10
kde-plasma/plasma-meta: add mediacenter and sdk useflags

Vindex17:
I think it's better this way because it's more modular (mediacenter it's not something everyone needs)

### DIFF
--- a/kde-plasma/plasma-meta/plasma-meta-5.4.49.9999.ebuild
+++ b/kde-plasma/plasma-meta/plasma-meta-5.4.49.9999.ebuild
@@ -12,7 +12,7 @@ HOMEPAGE="https://www.kde.org/workspaces/plasmadesktop/"
 LICENSE="metapackage"
 SLOT="5"
 KEYWORDS=""
-IUSE="bluetooth +display-manager gtk +sddm networkmanager pulseaudio +wallpapers"
+IUSE="bluetooth +display-manager gtk mediacenter networkmanager pulseaudio +sddm sdk +wallpapers"
 
 RDEPEND="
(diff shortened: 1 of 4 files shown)


kensington:
Thanks, this looks good. Just needs one small is needed: see check output:

    IUSE.invalid [fatal]          1

    kde-plasma/plasma-meta/plasma-meta-5.4.1.ebuild: mediacenter

Since mediacenter is not a global USE flag, it needs to be added to this package's metadata.xml.

Vindex17:
Ok. Just another couple of question before proceeding:
1) Do I have to apply it to git packages too?
2) Can I do the same thing with plasma-sdk dep?

kensington:
1. Yep, please apply to git packages too. We use these ebuilds as the basis for all version bumps, so any changes not applied there will end up lost.
2. OK with me.

kde-plasma/plasma-meta/plasma-meta-5.4.49.9999.ebuild
```diff
@@ -11,8 +11,8 @@ HOMEPAGE="https://www.kde.org/workspaces/plasmadesktop/"
 
 LICENSE="metapackage"
 SLOT="5"
-KEYWORDS=""
-IUSE="bluetooth +display-manager gtk +sddm networkmanager pulseaudio +wallpapers"
+KEYWORDS="~amd64"
```
kensington: Oops, keywords don't belong on live ebuilds.

kensington:
Thanks, rebased and merged as 782162b3ada9c1092f5b944380f698c2b420c668.

PR #375, opened 2015-09-13
x11-themes/gnome-breeze: Separate gtk and gtk3

awilfox:
This allows users to choose which version to install (or both) keeping
the system clean of any unneeded artefacts.

### DIFF
--- a/media-gfx/digikam/digikam-5.0.0_beta4.ebuild
+++ b/media-gfx/digikam/digikam-5.0.0_beta4.ebuild
@@ -75,7 +75,7 @@ COMMON_DEPEND="
 	)
 	scanner? ( $(add_kdeapps_dep libksane) )
 	gphoto2? ( media-libs/libgphoto2:= )
-	kipi? ( $(add_kdeapps_dep libkipi) )
+	kipi? ( $(add_kdeapps_dep libkipi '' '16.03.80') )
 	lensfun? ( media-libs/lensfun )
 	marble? (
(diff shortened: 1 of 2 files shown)


mrueg:
While I don't think its worth to split up the theme (as there are no additional deps pulled), as a user I probably want to have both active by default.

kensington:
repoman failed:
    IUSE.invalid [fatal]          1
    
    x11-themes/gnome-breeze/gnome-breeze-9999.ebuild: gtk3

so it needs a local USE flag description in metadata.xml

Should the other flag be called gtk2 instead of gtk? I'm not exactly sure what the usual handling of those flags is.

awilfox:
I haven't found any consistency in portage.  Some packages use gtk / gtk3.  Some package use gtk2 / gtk3.  A very few actually use gtk2 / gtk, and gtk means gtk3.

Maybe it should be doing gtk2 and gtk3 just to be completely unambiguous.  Also, sorry about that!  I can add gtk3 (and gtk2 if it isn't there) to metadata.xml.  I remembered to run repoman on the kate commit but not this one, doh.

mrueg:
I still feel that this is too much effort just to separate the themes. Someone else might want to install only the light or the dark theme, should we support that too?

a17r:
This doesn't concern me much as long as there is kde-plasma/kde-gtk-config which actually depends on both gtk+2 and gtk+3 unconditionally, so far.

PR #376, opened 2015-09-13
kde-apps/poxml: ANTLR hasn't been used for years so remove that stuff

chewi:
As mentioned on #gentoo-kde.


kensington:
Thanks, rebased as 11adc6ddd1710dac131a4536cea4fadcf95a08a4. I'll see if any other changes are required for the switch to gettext and then push to the main tree too.

PR #377, opened 2015-09-13
kate: Patch to workaround upstream #352627

awilfox:
This fixes a nasty bug that causes a segfault when opening files versioned with git when dev-libs/libgit2 is also installed on the machine.


mrueg:
Please add a more precise commit message. What does it fix?

awilfox:
I changed the commit message to be more descriptive.  I've forgotten git history serves as changelog now.  Sorry.

kensington:
Thanks, rebased as 57c2a20a71386e86ad4e49d5540b7e99ab28dd8d.

PR #378, opened 2015-09-13
kde-apps/kdepim{libs,-runtime}: Fix Super Evil KDE PIM Bug No. 1

a17r:
(hope it works for .1 as well, had to insert two newlines at end of files that were not present in .0 but also nowhere to be found in diff to .1)


kensington:
Thanks, rebased as bb093e3f89752911fd56abe94a75b125c5fa897f and b96e97c4cde08b064870110cb7b4edc314b05144.

PR #380, opened 2015-09-14
kde-apps/kdepim: Fix blocking UI call when opening attachment

a17r:
Upstream fix is in >=15.08.2
See also: http://bugs.kde.org/show_bug.cgi?id=350737


mrueg:
https://github.com/gentoo/kde/commit/e33ca366b83a445a3fe2cba9705a1ac306a2c3ee Thanks!

PR #382, opened 2015-09-15
kde-apps/kapture: New package

a17r:
Contrary to original plan, KScreengenie will not be renamed KSnapshot, instead it is to be called 'kapture' in the future.
Reference: https://www.mail-archive.com/kde-community@kde.org/msg01577.html


kensington:
Thanks, rebased as 79b1a7a1c77d3a16a307756945b1f47b1b76d956, 8a7d9412c8a7a9315f94e5916dc251719c675ae5, and f5b1fe8406faea81e5d0a786e36a51bba2bb6ca5.

PR #384, opened 2015-09-16
l10n: Bugfixes, missing version bumps

a17r:
Be more verbose about what is being copied or removed.
Merge everything from LEGACY_LANGS, not just messages.
Take care of nonexistent directories in destination paths.
Fixes https://bugs.gentoo.org/show_bug.cgi?id=560336

EDIT: Add missing version bumps.


mrueg:
Thanks, merged!

PR #393, opened 2015-09-19
net-misc/smb4k: Add block against net-misc/smb4k:4

a17r:


mrueg:
https://github.com/gentoo/kde/commit/621226c80adc163df72e40db825fafb3b3d1bd37 Thanks!

PR #396, opened 2015-09-20
app-text/kbibtex: Remove kdelibs4support DEPEND, add gcc check

a17r:
Package does not build using <gcc-4.9

### DIFF
--- a/app-text/kbibtex/kbibtex-9999.ebuild
+++ b/app-text/kbibtex/kbibtex-9999.ebuild
@@ -4,6 +4,7 @@
 
 EAPI=5
 
+KDE_GCC_MINIMAL="4.9"
 inherit kde5
 
 DESCRIPTION="BibTeX editor to edit bibliographies used with LaTeX"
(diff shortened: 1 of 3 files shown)


app-text/kbibtex/kbibtex-9999.ebuild
```diff
@@ -47,6 +46,15 @@ DEPEND="
 "
 RDEPEND="${DEPEND}
 	dev-tex/bibtex2html
+	!app-text/kbibtex:4
 "
 
 S=${WORKDIR}/${P/_/-}
+
+pkg_pretend() {
+	if [[ ${MERGE_TYPE} != binary  && $(tc-getCC) == *gcc* ]]; then
```
kensington: If I understand @mgorny correctly, the version check will work as expected with clang thus is not necessary to check for gcc.

kensington:
Thanks, rebased as 9bc9a46e2f7e358ad231fd2a1d0156efb2bb13e6, 4100b248d6d1d68111d33567c7bf09df76826985, and 061c734017f67a9ac2417563a9b485b270cd4719.

PR #414, opened 2015-10-05
Remove problematic blockers

Vindex17:
These changes fix unresolvable portage blocks. The transition of the kio-extras package from the former category is handled by the 3Q-2015 update file


kensington:
Thanks, rebased as ad1c936ed9b944c9d8da12da3a54d45f42431ed6.

PR #416, opened 2015-10-07
Add a live ebuild for the KF5 version of rsibreak

jktjkt:
I've changed the ${LICENSE} to show that it's actually GPLv2 or later
because that's what the source files say, based on a quick grep. There's
no support for translations so far because I have no idea how they work
in "kde5".

The blocker is in place because the eclass apparently only works for
kde-apps/, not kde-misc/.


kensington:
Thanks, rebased as 3c1878af2204277bdc84ac8e486c4af13fd35036 (with two small changes, added skeleton metadata.xml and prefixed commit message with $CPN).

PR #417, opened 2015-10-07
kde-apps/kdepimlibs:5 split

a17r:
Fix a few things while at it:

- KDE_DOXYGEN for libakonadi
- Add USE=tools to libakonadi (disables developer tool akonadi2xml and libxml dependency)
- KDE_DOC_DIR path for kdepim-kioslaves
- Fix DEPENDs here and there


kensington:
Thanks, rebased.

PR #424, opened 2015-10-10
kde-apps/kcalutils: Add missing DEPEND, drop unused

a17r:
Package now depends on grantlee


kensington:
Thanks, rebased as cafc86ca2794c2b12ed693959c3c38bd3f3449db.

PR #429, opened 2015-10-13
[app-office/kmymoney] Add kf5 version

a17r:


kensington:
Thanks, rebased as 165822b4f328c6e6f398079170f183ffcc37d3f3 and f4a377cf02be42518d8243e859b8a06521dd80e9 with updated commit message prefix.

a17r:
ah yes, commits were rather old already ;)

PR #430, opened 2015-10-14
kde-apps/kholidays: DEPENDs only on dev-qt/qtcore

a17r:
In preparation of becoming Tier 1 Framework.


kensington:
Thanks, rebased as a99b8642670862a718892f97108731b0cd40891d.

PR #431, opened 2015-10-14
kde-apps/kdebase-runtime-meta: Make more RDEPENDs optional

a17r:
kglobalaccel, kuiserver: Do it like upstream and don't RDEPEND if USE=minimal
See upstream commit be3ee5b9a36e1b7f3c6eab466e33762ae101b3c6

kwalletd: Make it optional with USE=oldwallet


kensington:
Thanks, rebased as dee669499b78692e054ce2ba07d40234b4c4f93f.

PR #432, opened 2015-10-14
media-plugins/kipi-plugins: Upstream removed DEPENDs

a17r:


kensington:
Thanks, rebased as cc8229f62c2852b43e5b8dcb81278ae7a09b463d.

PR #435, opened 2015-10-15
Adjust my email

Vindex17:


kensington:
Documentation/CONTRIBUTORS is auto-generated based on git log, so .mailmap would need updating or this change will be reverted when regenerate-files is next run.

Vindex17:
I'm not in .mailmap

Vindex17:
Ah ok, I think I figured out what you mean

kensington:
Thanks, applied.

PR #440, opened 2015-10-15
kde-apps/kpimtextedit: Fix DEPENDs

a17r:
composereditor-ng was merged into kpimtextedit.


kensington:
Thanks, rebased as e19fa7919dddd71fb6da141071b9b69824336627.

PR #444, opened 2015-10-18
kde-frameworks: Drop dependencies

a17r:
See also: https://git.reviewboard.kde.org/r/125673/

kwallet - not sure about use flag name - it is optional in a few places where so far "kde" has been used.


kensington:
Thanks, rebased as 50c71c2796a2daf93a76f57783e4f16ee7db6997 and 525ee6d049cff5324781fa80e3aad73ae67dd584.

PR #446, opened 2015-10-20
kde-apps/akonadi-search: Add new DEPENDs

a17r:


kensington:
Thanks, rebased as gnome-extra/zenity.

PR #448, opened 2015-10-20
kde-plasma/kwin: Support multimedia USE flag instead gstreamer

waltercool:
Not sure if this may be useful, I don't personally use gstreamer010, but someone may need it. QtMultimedia supports both gstreamer1.0 and gstreamer010 in their API, and kwin requires QtMultimedia with gstreamer for the video plugin. So basically here my idea is supporting gstreamer010 for the same implementation instead just avoiding it.


kensington:
Thanks, applied in cc771837b72e5724c209c1e1b1fd805a1d356862 with some very minor changes.

waltercool:
Makes sense, thanks

PR #450, opened 2015-10-22
Drop a few blockers, move more 15.08.2 packages to tree (drop old)

a17r:
Depends on https://github.com/gentoo/gentoo/pull/274


kensington:
Thanks, applied.

PR #459, opened 2015-11-04
media-video/kamoso: Removed xv for gst-plugins-meta dependency due incompatibility with gst-plugins-meta:1.0

waltercool:
Proposing removing xv use flag dependency because gst-plugins-meta:1.0 doesn't contain xv, this dependency was making as only alternative gst-plguins-meta:0.10 only.

Not sure how this can affect gstreamer 0.10, but this modification make it correctly works with gstreamer 1.6.0 with gst-plugins-meta-1.6.


kensington:
Thanks, rebased as 86e0134012f079c71be08cab870d9347c5263b65.

We could have used `xv(+)` for compatibility with 0.10, but judging by upstream commits it seems it's no longer needed there either.

PR #467, opened 2015-11-09
kde-plasma/plasma-workspace: Fix dead button in drkonqi

a17r:
Not nice: Getting a crash, then going through drkonqi and encountering a usability bug.


kensington:
Thanks!

PR #469, opened 2015-11-10
kde-frameworks/breeze-icons: Split out of kde-plasma/breeze-5.4.3

a17r:
...in before unmask? :)

Tested of course with 5.4.2, but I have no reason to believe it would not work in 5.4.3 (no changes to CMakeLists.txt).


kensington:
Thanks!

PR #470, opened 2015-11-10
kde-apps/oxygen-icons: pkgmove to kde-frameworks/oxygen-icons

a17r:
Should be coordinated with https://github.com/gentoo/gentoo/pull/345

Version bump to 5.16.0 -> can't do that myself


kensington:
Thanks!

PR #471, opened 2015-11-10
kde-apps: Add 15.08.3 release

a17r:
Added existing libkface fixes from tree.
Added USE=minimal to libkipi, from tree
Also contains gwenview libkipi-API fix taken from https://git.reviewboard.kde.org/r/125986/


kensington:
Thanks!

PR #473, opened 2015-11-12
[kde-misc/kdeconnect] Adjust to removed wayland USE flag of kwin

eliasp:
The 'wayland' USE flag of kwin was removed in b0ba1ce


kensington:
Thanks, rebased as 36d098a29eb4bd9a8f5bc67cbea6c500232546f7.

PR #475, opened 2015-11-13
kde-plasma/plasma-workspace: Updates

a17r:
btw, is there a reason to depend on sys-libs/pam instead of virtual/pam?


kensington:
virtual/pam should be fine, I forgot it even existed.

a17r:
This depends on https://github.com/gentoo/gentoo/pull/364 now.

kensington:
Thanks a lot! :)

PR #483, opened 2015-11-15
kde-apps/libkonq: Fix USE=minimal, depend on KF >=5.16

a17r:
Upstream does not install that file when built w/ KF 5.16 so hack can go away eventually.


kensington:
If we require 5.16, doesn't the USE="minimal"/src_install become a noop?

a17r:
Yes, but I figured it should be kept for a bit so people have a non-conflicting libkonq installed, before the block on libkonq[-minimal] can be removed from kde-frameworks/kio.

kensington:
Makes sense, thanks! Rebased as 5326546d73684ffa3453977d7dfe34952ea9a473.

PR #494, opened 2015-11-18
media-plugins/kipi-plugins: Fix DEPENDs, new USE

a17r:


kensington:
Thanks, rebased as ad07cf956d5578ae732590613dddfd9decb8a778.

PR #499, opened 2015-11-22
Move a bigger part of 15.08.3 to tree

a17r:
To be merged after https://github.com/gentoo/gentoo/pull/385


mgorny:
@gentoo/kde?

a17r:
kteatime currently does not link with KF5::Crash when it should actually use it (upstream bug)

kensington:
Thanks, rebased.

PR #504, opened 2015-11-30
kde-misc/akonadi-phabricator-resource: New package

a17r:
Pity it only seems to build with kdepimlibs-15.08.3 right now.

@kensington ? ;)


kensington:
phab!!! :D

PR #505, opened 2015-11-30
Fix bug #561470 by adding missing runtime dependency on net-misc/socat

fkoester:


kensington:
Thanks, rebased (applying to live ebuild too since that's what future bumps are based on) in a5216664de0767250003acc420317c67dc5b0041.

PR #508, opened 2015-12-03
kde-apps/konsole: Fix blocker, bug 567472

a17r:
See also: https://github.com/gentoo/gentoo/pull/433


kensington:
Thanks!

PR #510, opened 2015-12-05
kde-frameworks/frameworkintegration: plasma now needs noto fonts

Vindex17:
https://quickgit.kde.org/?p=frameworkintegration.git&a=commit&h=3ebe68904cf633ed2fbd4731588d5d4dce2a2b11

### DIFF
--- a/kde-frameworks/frameworkintegration/frameworkintegration-5.16.0-r1.ebuild
+++ b/kde-frameworks/frameworkintegration/frameworkintegration-5.16.0-r1.ebuild
@@ -4,12 +4,13 @@
 
 EAPI=5
 
+QT_MINIMAL="5.5.1"
 VIRTUALX_REQUIRED="test"
 inherit kde5
 
 DESCRIPTION="Framework for integrating Qt applications with KDE workspaces"
 LICENSE="LGPL-2+"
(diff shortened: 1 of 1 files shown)


kensington:
Thanks, rebased as 40f2055b133870d2324fbbe4bb7e3735958bd1f6. I'll fix 5.16 in the main tree separately.

shiba89:
Can this be made a useflag? Noto is ~500MB once installed and I don't even plan to use it.

kensington:
@shiba89 Put it in your package.provided. Upstream requires Noto by default and we try to keep as close to upstream as possible.

shiba89:
Seems reasonable. I'll do what you suggested, thank you very much.

kensington:
For a more permanent solution we could consider splitting media-fonts/noto into linguas-specific packages, but that would be up to @gentoo/fonts.

minlexx:
@kensington I heard only chinese or other asian fonts in noto collection consume so much space?

kensington:
@minlexx I think so. Adding some linguas USE flag(s) would allow for substantial reduction in installed space usage, but we might also consider repacking the tarball if download size is also a concern.

PR #513, opened 2015-12-09
Split kdepim, kdepimlibs: Fixes for 15.08, 15.12 and git

a17r:
Separate repositories for split kdepim libraries are happening any minute now.


kensington:
Thanks a lot, rebased and pushed!

PR #523, opened 2015-12-15
dev-util/kdev*: Fixes, workaround kapptemplate blocker

a17r:
The problem: A file was moved from kapptemplate - git master - to kdevelop, not going to be fixed in 15.12.
The fix: Add kdevelop 5.0 stable branch and remove the file there to have a compatible ebuild.

Also: Fixes all around.


kensington:
Thanks! Rebased.

PR #524, opened 2015-12-15
dev-util/kdevelop-valgrind: Fix DEPENDs, kdelibs4support was dropped

a17r:


kensington:
Thanks!

PR #528, opened 2015-12-19
{mail,pim}common: Add USE=designer, akonadi-search: fix xapian DEPEND

a17r:
See also: https://github.com/gentoo/gentoo/pull/493


kensington:
Thanks, rebased!

PR #530, opened 2015-12-20
kde-plasma/systemsettings: Add USE=classic, force handbook optional

a17r:
New use flag makes classic mode plugin optional.


kensington:
Thanks, rebased as 7dfb6d0f9e5795a0050ce6315236ce16bd805b92.

PR #536, opened 2015-12-27
kde-apps: Follow-up KDE PIM pkgmove in tree

a17r:
See also: https://github.com/gentoo/gentoo/pull/539


kensington:
Thanks, rebased as 4291d03f3d98ae1cf79788095d1932b541fa278d and 1ea92869fbb7afdff5df64adbdc1573ecb663a58.

PR #540, opened 2015-12-29
app-backup/kfoldersync: New package, bug 507500

a17r:
Based on work done by Frank Krömmelbein and Lorenzo Porta (Vindex17)
See also https://bugs.gentoo.org/show_bug.cgi?id=507500


Vindex17:
Why these?
> $(add_frameworks_dep kconfig)
> $(add_frameworks_dep kconfigwidgets)
> $(add_frameworks_dep kitemviews)
> $(add_frameworks_dep kwidgetsaddons)
> dev-qt/qtgui:5

a17r:
kfoldersync binary ends up linking with them.

Vindex17:
Because they are mandatory dependencies of the others, in fact I didn't even put them in CMakeLists.txt

app-backup/kfoldersync/kfoldersync-3.0.0.ebuild
```diff
@@ -0,0 +1,34 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI=5
+
+KDE_GCC_MINIMAL="4.9"
```
Vindex17: Thank you for pointing that out. I think I'll add a gcc version check even in CMakeLists.txt in the next release

kensington:
In fact all of those dependencies are directly used - check the header includes.

kensington:
Thanks, rebased as d3aff3cf218f5b259e9891473f4c189e7168daf2.

PR #541, opened 2015-12-29
kde4-base.eclass: Alt. kde-frameworks/oxygen-icons RDEPEND, bug 565736

a17r:


kensington:
Did this work OK for you? I worked on this a bit a while ago without luck, but that may well have been a dyndeps and/or invisible atom issue.

a17r:
Let me re-check that in the evening. On the other hand if it doesn't work, people can always put :5 into package.mask.

PR #542, opened 2015-12-30
kde-misc/plasma-applet-weather-widget: New package

a17r:
GHNS version does not support offline data cache.


kensington:
Thanks!

PR #546, opened 2016-01-02
kde5.eclass: In kde5_src_prepare, place cmake-utils_src_prepare on top

a17r:
This change makes PATCHES handling less error prone, as epatch will find
a pristine directory before there is punting and commenting.

Also tested possible negative impact with the following command in overlay dir:
``for file in $(find . -name "*ebuild" -exec grep -l PATCHES {} \;); do USE="-handbook" ebuild $file clean prepare; done``
...all good


kensington:
Thanks!

PR #548, opened 2016-01-03
kde-frameworks/kio: Drop obsolete blocker

a17r:
Enough time has passed since 5326546d73684ffa3453977d7dfe34952ea9a473
so no one should be on old and conflicting libkonq anymore.


kensington:
Thanks, picked as 67f8c45766ad9aa7d76d922d5530b71b36a3248e.

PR #549, opened 2016-01-04
RFC: kde-apps/konsolepart: New package; relax plasma-workspace-wallpapers blocker

a17r:
This should be given some testing. The current situation with konsole, where people end up without konsole binary after their switch to plasma profile + rebuild is less than ideal (happens for stable). I didn't really look at how other distros do the split - konsolepart is for the most part what had been konsole:4[minimal] minus desktop files that I think are strictly konsole binary related and with handbook disabled.

1. Add alternative dependency (where applicable?) on konsolepart in tree for the following packages:
 - app-editors/kile-2.1.3-r1
 - dev-util/kdevplatform-1.7.2:konsole
 - kde-apps/kdebase-meta-4.14.3*
 - kde-misc/krcstat-0.7.6
 - kde-misc/quadkonsole-0.4.4
 - kde-misc/yakuake-2.9.9-r2
 - net-analyzer/netwag-5.39.0 (doesn't look as if it actually needs konsolepart, just some terminal)
 - also add konsolepart dep to kde-apps/dolphin:4? It didn't even depend on konsole so far.
2. Move konsolepart to tree
3. Restore keywords for konsole-4.14.3-r2 and move to tree (bumped without so it causes no conflicts)
4. Ideally, stabilisation :)

A first, quick test on a pure KDE4 system revealed no issues with the upgrade.


kensington:
Thanks a lot, this will go a long way to helping avoid blockers.

PR #551, opened 2016-01-07
[RFC] kde5-functions.eclass: Introduce add_qt_dep for consistent Qt version deps

a17r:
I guess it's not a new idea. ;)

We can't expect portage to properly resolve Qt version dependencies by just setting >=dev-qt/qt{core,test}-${QT_MINIMAL} in eclass and rely on dev-qt/*'s own version blockers for the rest.

This also enables global QT_MINIMAL settings if multiple Plasma versions are around that might be better off with different Qt versions.

In the end will also solve https://bugs.gentoo.org/show_bug.cgi?id=565764

For now I've converted ebuilds where QT_MINIMAL had already been in use.

### DIFF
--- a/app-office/kexi/kexi-9999.ebuild
+++ b/app-office/kexi/kexi-9999.ebuild
@@ -33,12 +33,12 @@ COMMON_DEPEND="
 	$(add_kdeapps_dep kdb 'mysql?,postgres?,sqlite?')
 	$(add_kdeapps_dep kproperty)
 	$(add_kdeapps_dep kreport)
-	dev-qt/designer:5
-	dev-qt/qtgui:5
-	dev-qt/qtnetwork:5
-	dev-qt/qtprintsupport:5
-	dev-qt/qtwidgets:5
-	dev-qt/qtxml:5
+	$(add_qt_dep designer)
+	$(add_qt_dep qtgui)
+	$(add_qt_dep qtnetwork)
+	$(add_qt_dep qtprintsupport)
+	$(add_qt_dep qtwidgets)
+	$(add_qt_dep qtxml)
 	activities? ( $(add_frameworks_dep kactivities) )
 	marble? ( $(add_kdeapps_dep marble) )
 	mdb? ( dev-libs/glib:2 )
@@ -48,7 +48,7 @@ COMMON_DEPEND="
 		dev-libs/libpqxx
 	)
 	sybase? ( dev-db/freetds )
-	webform? ( dev-qt/qtwebkit:5 )
+	webform? ( $(add_qt_dep qtwebkit) )
 	xbase? ( dev-db/xbase )
 "
 DEPEND="${COMMON_DEPEND}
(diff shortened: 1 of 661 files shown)


a17r:
See also: https://forums.gentoo.org/viewtopic-t-1037026.html

kensington:
Well the mixed versions bugs are a problem with the qt ebuilds and will be fixed in 5.6.0 at the latest (with a plan to fix in 5.5.1)

a17r:
whatever works ;)

kensington:
@Pesa what do you think?

Pesa:
Well I'm not familiar with the kde stuff, but if the only problem is the mixing of different Qt versions, then it should be solved by the recent revbumps (5.5.1-r1). If there's something else, please elaborate.

a17r:
The problem until the qt revbumps has been that kde packages gave mixed version deps, like:
>=dev-qt/qt{core,test}-5.5.1 (by eclass)
dev-qt/qt*

For stable users, that broke --autounmask-write and they ended up with useless package.keywords entries.

Pesa:
I'm not sure I understood your comment. Are you saying it's fixed now? so can we reject this PR?

a17r:
Sorry for not being clear; I just assumed it will work like with Qt4 now, but I couldn't test it so far.

a17r:
OK, I guess the revbumps did fix the issue with e.g. designer installed at 5.4.2 with the rest of qt at 5.5.1 - but the problem coming from KDE packages defining only qtcore-${QT_MINIMAL} is still there, since the blockers don't help portage figure out the fix (pull in all with compatible versions only).

This is how it looks for a stable user: https://bpaste.net/show/18db312735d1

Pesa:
Ok that's definitely an issue with portage's dependency solver being unable to provide a reasonable suggestion when a valid solution cannot be found. I suppose that simplifying its job by listing more precise (stricter) deps would help. Can you give a sample of the output after your patch?

a17r:
This is how it looks with the patch: https://paste.kde.org/plye61et7

Note that it still won't complete because only those few ebuilds have been converted so far. However the qt packages pulled in by lokalize (qtdbus, gui, script, sql, widgets, xml) are consistently at 5.5.1.

Pesa:
Yep, it does indeed look better. portage is able to suggest several more (correct) keyword changes with the patch.

So yes it's definitely an improvement. However as you said it requires converting a lot of consumers in order to be fully effective, and in some cases just one unconverted package is enough to spoil the party. Sounds like it could be a lot of work for little(?) gain...

a17r:
Yes, definitely wouldn't convert manually ;) But a little script should do the trick - I think it would improve the situation for probably 95% of Plasma-5 upgraders, and makes the blocks easier to solve for the rest.

a17r:
TODO: Support for defining subslot

a17r:
...not sure if it shall be a subslot operator only or full slot override

a17r:
stable user's plasma-meta upgrade with current state of converted deps: https://paste.kde.org/pqqgn5u1c

No mention of Qt 5.4.2 anymore (KDE blocks are there because this system is not up for migration yet).

kensington:
Thanks, picked as 17b435e617b0a349ed6de0b37850b8f80bf432f2...fcbbbe91526ebbbea466398d35e8636bdfaa7ff3.

PR #552, opened 2016-01-07
kde-apps/konsole: Auto-block slot 4 again

a17r:
Same should be done for kde-apps/konsole-15.08.3 in tree...


kensington:
Thanks, rebased as 4d8dacd2e270ba3fb66e424e21e3cad4792236e8 and 129d0bb8d550114379c3d9d4c9bbbb63b2d5d205.

PR #558, opened 2016-01-12
kde-apps: Applications 15.12.1 version bump

a17r:
Dropped .0 because I wanted to see any differences to .1 - however don't see a reason to keep old on overlay any longer.


mrueg:
Merged, thanks!

PR #559, opened 2016-01-17
kde-frameworks/polkit-qt: New package

a17r:
Wait for final decision upstream or merge already?
Will need some dependency adjustments in tree as well before it can be used by everyone.


kensington:
Thanks a lot!

PR #563, opened 2016-01-18
kde-apps/kwalletd-pam: Add aptly named slot 4 version of kwallet(d)-pam

a17r:
Separate package for use with KDE 4 or Plasma-5 with oldwallet still present. It could be done with multibuild as well and some use flag, but this way it could be made a dependency of KDE 4 without having to care about possible future ties into kf5.


kensington:
Thanks, rebased as 754f8c9d426cdf010e3a89bf7775b8dc79432568.

PR #564, opened 2016-01-18
kde-apps/kdecore-meta: New package

a17r:
This makes kdebase-meta stand aside of kde-apps-meta, while keeping dolphin, konsole and kwrite pulled in.

Makes kde-apps-meta not pull in plasma-meta, in theory.


kensington:
Thanks!

PR #577, opened 2016-01-27
kde-apps/{kdesdk-strigi-analyzer,mplayerthumbs}: Remove neglected upstream packages

a17r:
No code change since 4.10 (save for some minor translation bits) and doesn't work with dolphin:5, only serves to confuse people.

I'd even go as far as drop 15.08.3 in tree before it is stabilised, but oh well.


kensington:
Propose upstream to drop it too?

a17r:
Let's see if someone reacts @upstream, I'd say dropping it downstream is a minor incident as long as the full KDE Applications is not even side-by-side packageable by default.

kensington:
Thanks, picked as 718ea508b5eb46b9c48bfb8af5178920f735ea34 and 6d3ab442b506ad899ebee38d324dbd3aca587ca8. Added a note on the wiki too to follow up upstream.

PR #578, opened 2016-01-29
sys-block/partitionmanager: kf5-based kdesu workaround is unnecessary

stikonas:
partitionmanager is now able to locate libexec path directly.


kensington:
Thanks, picked as 31b7df03bc8231c6c4f0e862ac806c16d70c96ec.

PR #579, opened 2016-01-30
kde-apps/libkface, media-gfx/digikam: Add subslot operator

a17r:
In anticipation of opencv-3.1 subslot bump

See also: https://github.com/gentoo/gentoo/pull/752


kensington:
Thanks!

PR #584, opened 2016-02-03
kde-apps: Add KDE PIM 4.14.11 snapshot

a17r:
Changes:
- ~~Drop gpgmepp build from kdepimlibs, RDEPEND on standalone gpgmepp:4 package instead~~
- ~~^ one less blocker~~
- Drop USE="qt4 qt5" from akonadi-server and make it qt4-only again
- Remove USE=facebook from kdepim-runtime (API dropped)
- Fixes bug 560376 ("multiple merge candidates, aborting")
- Fixes build with newer boost (bug 556104) without a huge patch
- No further kdepim-icons package, alternatively depend on kde-frameworks/oxygen-icons


kensington:
Thanks, picked as d619f03c7796857a713e5c9e7bb81b17e71f7341...dc3ed9c3dbcf7336d12e04723b97255518ce77ff.

PR #585, opened 2016-02-05
kde-plasma/kscreenlocker: Fix blocker, bug 573890

a17r:
Should go into tree as well...


mrueg:
Thanks! Applied as 9ec6d0262ee0c92aac31522335d3d76e7400c5f1

PR #586, opened 2016-02-07
kde-frameworks: Add 5.19.0 release

a17r:
A little bit too early, but why throw it away... :)


a17r:
oh, of course - missing digest for lack of access...

mrueg:
How about doing your quizzes instead? That would give you access ;)

kensington:
I'll merge this once I finish squashing EAPI 6 porting on the top.

a17r:
I'd also like to add one patch that sadly didn't make it into kactivities 5.19 anymore.

kensington:
btw, all in-tree versions should have ~arm so the keywording part of the script is broken (if you're using it, that is).

kensington:
Should all now be part of b912038f260872682bec207f6153a26956949dc5...dfe031ff1801e8aee728ebdcccfa7c3ab7df9368.

PR #590, opened 2016-02-12
kde-plasma/kwin: Add blocker against <kde-apps/kdeartwork-meta-15.08.3-r1

a17r:
It improves upgrade path from KDE SC 4. Since the Plasma-5 upgrade guide
removes all Plasma-4 depending packages from world, but kdeartwork-meta
is not necessarily part of depgraph on the subsequent rebuild, people
currently face a block if they still have kdeartwork-meta-4.14.3 and
kdeartwork-styles installed.


kensington:
Thanks! Picked as 14b9915d6169ac24ae5ae0326b6c91c078c6a076.

PR #592, opened 2016-02-14
kde-frameworks/oxygen-icons: Add blocker against conflicting kdepim:5

a17r:
...plus unblocking solution for 15.12 branch.

Slow system crawling through upgrade right now. ...built fine.


kensington:
Thanks!

PR #593, opened 2016-02-15
kde-apps/kdepim: Some splitting / KDE PIM 15.12.2 bump

a17r:
kdepim-meta-15.12.2 built fine ;)

New package:
- libktnef ('soft' move from ktnef:5 - solved by 15.12.2 bump, only live users might get a conflict)

kde5.eclass:
- Add new vars for split PIM handling
- Check for KMNAME=kdepim in src_prepare and bulk-edit dependencies

New (old) split KDE PIM packages:
- ktnef
- akonadiconsole
- blogilo
- calendarjanitor
- konsolekalendar
- kaddressbook
- knotes
- akregator


kensington:
We could potentially do something like what we did in the past with kde4-meta - ``[[ ${KMNAME} = kdepim ]] && sed -i -e "/find_package/s/REQUIRED//"`` to punt all the bogus deps.

kensington:
Thanks!

PR #596, opened 2016-02-18
kde-apps: Applications 15.12.2 version bump (minus PIM)

a17r:
This is actually just about the last three commits, just based on https://github.com/gentoo/kde/pull/593 (also depends on the package.mask file from there)

Unlike the PIM bump, this one is not yet build tested.


kensington:
Thanks!

PR #603, opened 2016-02-21
Add kwallet(d)-pam RDEPENDs

a17r:
That's the first part, second part of it will need to go into login manager's pam.d configs.


Arfrever:
I think that these new dependencies should be conditional on pam USE flag.

a17r:
One more optional runtime dependency, anyone? ;)

Arfrever:
AFAIK KDE 4 can be installed on system without PAM installed at all.

In case of KDE 5, dependency is optional in kde-plasma/kscreenlocker, but unnecessarily unconditional in kde-plasma/plasma-meta. So kde-plasma/plasma-meta should be fixed.

a17r:
One alternative I can think of is putting kwallet-pam rdepend one level above into plasma-meta[pam] and inside whatever meta package that gets kwallet:4 pulled in, and have it mentioned in kwallet as a postinst message.

kensington:
Really nice work, thanks! Picked as 220df2a284168cbfaeb5028b826461421467c2e4...591a6dd170e989be53af3e6e7fc55be53a6c7695.

PR #623, opened 2016-03-10
[WIP] kde-apps: Add USE=+qt4 to meta ebuilds

a17r:
Add USE=qt4 to control install of KDE4-based applications through meta ebuilds.

USE=pim-conflict: not sure if serious...


johu:
:-1: I don't like the over complification

PR #624, opened 2016-03-11
plasma-nm needs new enough QCA due to -DQT_NO_KEYWORDS

jktjkt:
plasma-nm started building with -DQT_NO_KEYWORDS which removes Qt's
"signals" and "slots" keywords, leaving only Q_SIGNALS and Q_SLOTS in
place. This has happened in plasma-nm's commit 705d0f2f.

QCA got fixed to not require the signals/slots "keywords" in 66b9754,
which is only part of the v2.1.1 git tag.

I have no idea how the QCA tarballs are genrated of course because some
tags are apparently missing form that repo...


kensington:
Thanks! Picked as e393992f655309128bb62dfaa8d6d1b392e45ec4.

PR #626, opened 2016-03-13
Frameworks 5.20 was released - more conflicts!

a17r:
https://www.mail-archive.com/kde-core-devel@kde.org/msg20595.html

Added transitional packages for Plasma 5.5 so that it does not depend on unreleased Plasma 5.6.


johu:
You are adding a package with "arm". This is not possible without involving arch team.

a17r:
Not sure what to do then to be able to release KF-5.20 without breaking ~arm with Plasma 5.5.5. kactivities-workspace consists of modules that have been part of KF until 5.19, and will be only released with Plasma 5.6.

johu:
You will need to do your quizzez to know that.

kensington:
Thanks, picked as a4067bc7d43134009dea39e19811553b282bfd0c...4120d94fbecaf44ba78ca7d1b4541807ca415d18.

For the record, I build the new packages on arm.

PR #627, opened 2016-03-13
RFC dev-libs/libunity: added 7.1.4

karolherbst:
when libunity is installed the taskmanager entries of chromium (and most
likely other applications) get additional information. For chromium that is
download progress and number of active downloads

This ebuild is trivially copied from the unity overlay and needs additional
work. I just wanted to ask you how we should deal with the ubuntu versioning
scheme.

I will tidy up the ebuild later, but you can play with it already


johu:
I would suggest to find an other maintainer. This is not realy kde only related.

mrueg:
@karolherbst  interested in proxy-maintaing it via @gentoo/proxy-maint  ?

karolherbst:
@mrueg can't say it yet, might be too busy in the near future

a17r:
Treecleaning of dev-libs/dee in https://bugs.gentoo.org/show_bug.cgi?id=603562 doesn't improve chances of this entering tree, unless a newer release does not depend on it...

karolherbst:
sad... no idea if by now there are other way to get those progess bars eveywhere. If yes, then this could be dropped.

PR #632, opened 2016-03-15
kde-apps: Drop some 15.12.3 ebuilds that moved to tree

a17r:
see also: https://github.com/gentoo/gentoo/pull/1064

Keeps the following kde-apps-meta-15.12.3 revisions in overlay:
-r3 pulls in kdeedu-meta:5, to be moved after final review
-r4 continues to pull in kdepim-meta:5 for overlay users

Keeps the following kdenetwork-meta-15.12.3 revision in overlay:
-r2 pulls in plasma-telepathy-meta:5, to be moved after final review

kdwebdev has no code change in months/years, so I would not bother with moving 15.12 to tree.


kensington:
Thanks, picked as 3aac060d6219e192c6f052a8433ea398d03fb9b0...1a1abff9b23350710e8adbdd3375e0ff2565b0c0.

PR #634, opened 2016-03-17
kde-frameworks/kio: Fix crash on in-place copy/paste

a17r:
See also: https://bugs.kde.org/show_bug.cgi?id=360488


kensington:
Thanks, picked as 9c0c4a346dfbe1d155e31d7ad0533c4e798decc9.

PR #641, opened 2016-03-29
kde-apps: Add KDE Applications 16.03.90

a17r:
New addition: LINGUAS=ast


johu:
Reverting seems a little bit odd, when you can just add the metadata.xml in an other commit.

a17r:
sure, I can do that as well

a17r:
re-worked much of kde-l10n +friends so that it only unpacks the required 4 or 5 part and is much faster in src_prepare

kensington:
Thanks, picked as 7952f3f7375d511d262b1603c1eb03e762f0b55f...98da7b34c0c01d6fbd2ad6463a42d5f7727587dd.

PR #647, opened 2016-04-03
kde-apps/kalzium: Add kf5-based ebuild

a17r:


kde-apps/kalzium/kalzium-5.9999.ebuild
```diff
@@ -0,0 +1,67 @@
...
+src_configure(){
+	# Fix missing finite()
+	[[ ${CHOST} == *-solaris* ]] && append-cppflags -DHAVE_IEEEFP_H
+
+	local mycmakeargs=(
+		-DCMAKE_DISABLE_FIND_PACKAGE_Eigen3=$(usex editor)
+		-DCMAKE_DISABLE_FIND_PACKAGE_AvogadroLibs=$(usex editor)
+		-DCMAKE_DISABLE_FIND_PACKAGE_OpenBabel2=$(usex editor)
+		-DCMAKE_DISABLE_FIND_PACKAGE_OCaml=$(usex solver)
+		-DCMAKE_DISABLE_FIND_PACKAGE_Libfacile=$(usex solver)
```
johu: cmake-utils_use_find_package is still available in EAPI 6

a17r: Yes I know, am I required to use it here? This was just the fastest way to convert from -DWITH_*

johu: It is more readable, that is the reason why it was not removed with EAPI 6. We should not mix ebuild style in kde project.

johu:
Thanks rebased onto kde overlay master.

PR #648, opened 2016-04-04
Fix typo in k3b-9999.ebuild

tsjk:


johu:
Thanks. Committed as 968b94c7875864b045393e26d870724fb386f9a7.

PR #649, opened 2016-04-04
kde-plasma/khelpcenter: pkgmove to kde-apps/khelpcenter

a17r:
Depends on: https://github.com/gentoo/gentoo/pull/1195

Here's what happens with khelpcenter translations:
* <khelpcenter-5.6.50: ships own khelpcenter.po (removed by PR)
* >=khelpcenter-16.03.90: relies on kde-l10n for translation
* >=kde-l10n-16.03.90 ships khelpcenter.po
* kde-l10n-15.12.3 has no khelpcenter.po (added via khelpcenter tarball by PR)

docs do not conflict, so as a result all kde-plasma and kde-applications releases can be combined after the changes.


kde-apps/khelpcenter/khelpcenter-5.6.49.9999.ebuild
```diff
@@ -9,6 +9,13 @@ inherit kde5
 
 DESCRIPTION="The KDE Help Center"
 HOMEPAGE+=" https://userbase.kde.org/KHelpCenter"
+if [[ ${KDE_BUILD_TYPE} = live ]] ; then
+	EGIT_REPO_URI="git://anongit.kde.org/${PN}"
```
johu: EGIT_REPO_URI is set by eclass already

johu:
khelpcenter-5.6.49.9999.ebuild can be dropped completly then...

a17r:
true, khelpcenter-16.04.x is preferable as soon as it arrives since it solves some major bug with search, something that should work in a khelpcenter

kde-plasma/plasma-meta/plasma-meta-5.6.49.9999.ebuild
```diff
@@ -53,7 +53,7 @@ RDEPEND="
 		$(add_plasma_dep breeze-gtk)
 		$(add_plasma_dep kde-gtk-config)
 	)
-	handbook? ( || ( $(add_kdeapps_dep khelpcenter) $(add_plasma_dep khelpcenter) ) )
+	handbook? ( $(add_kdeapps_dep khelpcenter '' ${PV}) )
```
johu: instead of PV use a constant release version

johu:
Commit "kde-base/kdelibs: Set SRC_URI in ebuild" has nothing to do with this PR

kde-plasma/plasma-meta/plasma-meta-5.6.49.9999.ebuild
a17r: well in this case it is PV so that kde-apps/khelpcenter dependency can correspond to 5.6.x instead of KDE_APPS_MINIMAL which would otherwise be 14.12.

johu: Then use 5.6.2

johu:
I would name the commit "kde-plasma/khelpcenter: pkgmove to kde-apps/khelpcenter" just "kde-plasma/khelpcenter -> kde-apps/khelpcenter".

profiles/updates/2Q-2016
```diff
@@ -0,0 +1,2 @@
+move kde-plasma/khelpcenter kde-apps/khelpcenter
+move kde-base/khelpcenter kde-apps/khelpcenter
```
johu: The second line is not needed.

a17r: I did want to be extra careful there since there still happen to be users around that installed 4.14.3 back when it was in the overlay. Remember very strange blockers that occurred because a pkgmove was only done in tree but didn't update /var/db/pkg for overlay users...

johu:
khelpcenter-5.6.49.9999.ebuild can be dropped

johu:
Thanks.  Picked a929f20d8736a062d52645a5a4f48fd6e17e4522...6e0480d70a4e0af373a45bf5d2f195699a4d259d

PR #654, opened 2016-04-07
kde-apps/libksane: Add block against libksane:4[-minimal]  | Fix kdepimlibs split pkgs

a17r:
See also: https://github.com/gentoo/gentoo/pull/1219


kensington:
Thanks!

PR #655, opened 2016-04-09
kde-apps/kde4-l10n: Don't die on empty directories

a17r:
Adds einfo about files not found. Use to update lists.


kensington:
Thanks a lot! Picked as 83b49bc08fe5dc565ffb404427f34b039e25cc18...931d6a9e8afd45422069cac3fdba223a455831db.

PR #656, opened 2016-04-10
kde-apps: Remove KDE PIM 15.12.3

a17r:
16.03.90 is neither better nor worse than 15.12.3, at least it is prettier when it fails.


kensington:
Thanks, picked as 0603dc5db3a5fd141bc6be8f81f7d98b58daae15...acde761d86573cb71bdf3286065d4db6c1a9206f.

PR #663, opened 2016-04-16
add KDevelop 5.0 Beta 2 public release

whitslack:
I have been using [KDevelop 5.0 Beta 2](https://www.kdevelop.org/news/kdevelop-50-beta2-release) for serious work for a few days and have found it to be at least as stable as KDevelop 4.7.3. I am contributing these ebuilds so that others may enjoy this fine software without needing to use the "live" versions, which are presently the only ones available in the KDE overlay.

These ebuilds are copied from the 5.0.9999 ebuilds with only trivial alterations. I have enhanced `kde5.eclass` to support guessing `SRC_URI` based on `${KDEBASE}`.

I have successfully built, installed, and used `dev-util/kdevplatform-4.90.91` and `dev-util/kdevelop-4.90.91` from these ebuilds on an `~amd64` system running Qt 5.5.1 and KDE Frameworks 5.21.0. I have successfully built and installed (but not used) the other three: `dev-util/kdevelop-pg-qt-1.90.92`, `dev-util/kdevelop-php-4.90.91`, and `dev-util/kdevelop-python-4.90.91`.


a17r:
Thanks! Are there any changes to the ebuilds? These should be also carried over to 9999 then.

whitslack:
@a17r Aside from setting `KEYWORDS="~amd64"` and unsetting `EGIT_BRANCH` (neither of which we would want to do in the live ebuilds), the only other change I made was to move `RESTRICT="test"` *above* `inherit kde5` in `dev-util/kdevelop` and `dev-util/kdevelop-python`. I did this because `kde5.eclass` appends `RESTRICT+=" mirror"` in case of unstable releases.

johu:
@whitslack RESTRICT should work below inherit statements, if this is a problem in eclass it should be fixed there first

whitslack:
@johu I agree: ideally it should be possible to define `RESTRICT` anywhere in the ebuild and for an inherited eclass to modify the defined value of `RESTRICT`, regardless of the relative ordering of the definition and the `inherit` call. However, I don't know enough about ebuild programming to achieve this.

```
do_the_inherit() {
	FOO+=" baz"
}

do_the_inherit
FOO="bar"
```

In the preceding example, how can `do_the_inherit()` be written so that `FOO` ends up with the value `bar baz` at the end of the script? I worked around the problem essentially by moving `FOO="bar"` before `do_the_inherit`, but you're telling me this should not be necessary. So what would you do?

johu:
@kensington i am just wondering: do we lost the RESTRICTs defined in eclass when we set the var in ebuild in past and didn't notice it?

kensington:
Correct, ``RESTRICT`` defined in an eclass does not stack the same was ``IUSE`` or ``DEPEND`` does.

whitslack:
@johu So now that we've determined that my movement of `RESTRICT=` above the `inherit` line was proper, what is remaining to get this merged?

whitslack:
(Rebased to incorporate 5057416b4657c988a4def75279816c453df5d957.)

johu:
Thanks. Could you please split eclass changes into a separate commit? And try to get the release ebuilds in sync with the live ebuilds by using '+='. This should cover to not overwrite the eclass values.

johu:
Thanks, picked.

PR #667, opened 2016-04-16
kde-apps/kreport: Python required for configure, fix deps

a17r:


kensington:
If it's build-time-only, should it be python-any-r1?

a17r:
thx, you're right ;)

kensington:
Thanks, picked as 61e891a0601897b3dc5a3703e1b33d2d07246af7...b847aa2914cad0b96b3f463d4bcbfa373edb44d4.

PR #670, opened 2016-04-23
kde-apps: KDE PIM depends on >=Qt-5.6.0 and qtwebengine

a17r:


johu:
We can merge this when Qt 5.6 is visible to kde overlay.

johu:
Thanks, picked.

PR #672, opened 2016-04-24
kde-apps KDE PIM: Split building upstream

a17r:
Pushing improvements for existing split ebuilds so kdepim live keeps building, while more split ebuilds are in the works.


kensington:
Thanks!

PR #673, opened 2016-04-25
sets: Drop non-existent dep (kde4-l10n-9999)

eliasp:


johu:
Thanks, picked as 70cff03a40a142fe8d43827e5245de786ffc72dc (reworded commit message).

PR #674, opened 2016-04-25
kde5.eclass: Fix recent change wrt split PIM 16.04.x

a17r:
Gentoo-bug: 581176


johu:
Thanks, picked as 52088d62d0545e436fc86de216dac07438ec82f1.

PR #677, opened 2016-05-02
kde-apps/messagelib: Block conflicting kdepim-addons

a17r:
Upstream fix for missing headers without kdepim-addons causes
file collision with the latter, see also:

https://bugs.kde.org/show_bug.cgi?id=362100
https://quickgit.kde.org/?p=messagelib.git&a=commit&h=d3c47f200c96e04785d05af9fb086bcd5a68f251


johu:
Thanks, picked as 626da49fc643211ccc0eefeb8c98d6cc02207698

PR #678, opened 2016-05-05
kde-apps/ark: Fix (some) tests, add missing DEPEND

a17r:
Gentoo-bug: 560548


johu:
If there are still failing tests we should restrict them.

a17r:
it's a sandbox error, might be solvable but didn't have time to look into it

kensington:
Thanks, picked as 171b6b4b75561a720c0d670561e94207401ba6ad.

PR #681, opened 2016-05-12
kde-apps: Add 16.04.1

a17r:


johu:
Thanks picked and pushed.

PR #683, opened 2016-05-14
dev-util/kdevelop-embedded: New package | some fixes

a17r:


johu:
Thanks. Picked and pushed. Please try to avoid mixing topics in the same PR.

PR #684, opened 2016-05-15
kde-apps/kde4-l10n: Set +minimal by default | cleanup blockers

a17r:
Eases use of non-plasma profile.


johu:
Thanks picked and pushed.

PR #688, opened 2016-05-19
kde-apps: Update blockers

a17r:
follow-up to https://github.com/gentoo/gentoo/pull/1470


johu:
Thanks. Picked and pushed.

PR #689, opened 2016-05-19
kde-apps: Drop per-package FRAMEWORKS_MINIMAL

a17r:
5.21.0 is now global minimum.


johu:
Thanks. Picked and pushed.

PR #692, opened 2016-05-29
Documentation: Remove Qt unmask from kde-frameworks-live.base

a17r:
See also: https://bugs.gentoo.org/show_bug.cgi?id=584448


Documentation/maintainers/regenerate-files
```diff
@@ -10,7 +10,7 @@ SCRIPT=$(basename $0)
 HEADER="# Autogenerated by ${SCRIPT}, DO NOT EDIT."
 
 KF_RELEASES="5.21 5.22"
-KF_RELEASE=${KF_RELEASE:-5.21}
+KF_RELEASE=${KF_RELEASE:-5.22}
```
johu: KDE Framworks 5.21 is stable 5.22 not. So why should the default sets be the unstable version?

johu:
Thanks. Squashed and rebased as 7c76d77281e05edf3f7183fd31cc98a29583c995. I have re-added kf-env to live.

PR #695, opened 2016-06-09
kde5-functions.eclass: Improve punt_bogus_dep

a17r:
The variable holding the min version requirement can be named anything.
This should catch most cases, most importantly Qt5 modules.

Fixes akonadi-9999 at least.


kensington:
Thanks!

PR #708, opened 2016-07-03
kde-apps/kdepim-l10n: Sync LINGUAS->L10N with tree

a17r:


johu:
Thanks. Picked  as 78565ea3fdb167fa375e352f70ad5eb22c0318a4.

PR #709, opened 2016-07-03
kde-plasma/kinfocenter: Drop USE="nfs,samba", add pkg_postinst messages

a17r:
Having seen some optional runtime deps in plasma-desktop and systemsettings, I thought why not put them in a separate package to avoid needless rebuilds.

For minimal interruptions this is done by
* keeping the old USE flags in 5.7.0 intact to gently enforce them in the new package (autounmask works)
* removing the USE flags in 9999 when for next version autounmask will have done its magic

In kde-plasma/kinfocenter the (imo) useless flags were just replaced with postinst messages.


johu:
👎 As already expressed my veto on irc: Imho this change does not improve anything accept a little bit compile time. It makes the  situation worse for current installed packages. The optional runtime dependencies should be coupled to the packages where they belong.

a17r:
Would you pick the kinfocenter change?

johu:
@kensington please add your opinion

kensington:
Thanks, picked as b61d5211e9565557ad067c3c2042ad3a807569a9

PR #711, opened 2016-07-04
Documentation: Add missing Qt-5.6.1 bits and regenerate files

a17r:


johu:
What about the rest of 5.6.1 in dev-qt/? Qt should be updated properly for other consumers.

a17r:
Could be added as well of course, this was added at first based on the old Frameworks more-deps list and now updated according to Plasma-5.6.5 autounmask needs.

Maybe just add a simple line ``dev-qt/*:5/5.6`` instead of the whole bunch?

johu:
If the regenerate script can handle this i am fine with the proposed change.

johu:
Thanks, picked as e00038914ab7123b4df5a2252ffb18038374b3f8.

PR #713, opened 2016-07-07
Some improvements

Vindex17:
- Pull in translations for kde applications.

- Powerdevil now has NetworkManager-related packages as hard dependencies, nullifying the effect of the networkmanager useflag. It's also not that useful on systems without batteries.

### DIFF
--- a/eclass/kde5.eclass
+++ b/eclass/kde5.eclass
@@ -179,8 +179,11 @@ case ${KDE_AUTODEPS} in
 			"
 		fi
 
-		if [[ ${KDE_BLOCK_SLOT4} = true && ${CATEGORY} = kde-apps ]]; then
-			RDEPEND+=" !kde-apps/${PN}:4"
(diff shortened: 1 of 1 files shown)


kensington:
powerdevil's networkmanager-qt dependency is still optional.

kensington:
We should discuss the l10n change as historically we're always required that be installed manually.

Vindex17:
Sorry, I just saw the cmake file. Another question: why don't use the global 'wifi' useflag then? (instead of 'wireless')

a17r:
It's for both wifi and bluetooth power save

kensington:
Thanks for your work, however after discussing it I'm afraid we have decided to decline this particular change as it will result in kde-apps/kde-l10n being unnecessarily pulled in for many users.

This is something we can revisit if runtime-switchable USE flags ever get implemented in a future EAPI.

PR #714, opened 2016-07-08
kde-apps/kde4-l10n: Make USE=minimal a no-op

a17r:


kensington:
Thanks, rebased as d55d4b5fb1fee6688755304c64dc5ee2bfa9bf57...26b81346c24b1c1833c468adc05f0331a9270df8.

PR #716, opened 2016-07-08
kde-apps/kde-apps-meta: Drop USE=minimal requirement on kde-apps/kde4-l10n

a17r:


kensington:
Thanks, bye bye minimal!

PR #718, opened 2016-07-08
KDE PIM: Fix LICENSEs

a17r:


johu:
I would squash the kde-apps commit to one.

johu:
Thanks, picked.

PR #720, opened 2016-07-10
kde5.eclass: Add kde*-l10n handling with KDE_L10N, make 3 new sr variants available

a17r:
updated kde4-l10n-16.04.2-remove-dirs may be synchronised to tree
btw this also fixes @mgorny 's fear of PMS violation :)


johu:
I am against to move the KDE_L10N linguas definitions for releases to the eclass, as this means we need to touch them on every release.

a17r:
hmm not necessarily, when 16.04.2 leaves the tree I might as well change the version check so that it covers all of 16.xx.x, but sure, can be moved into the ebuilds.

johu:
Thanks, picked. 

(dear god please make this l10n stuff ends... :D)

PR #721, opened 2016-07-12
kde-apps/{kleopatra,libkleo}: Fix DEPENDs

a17r:
~~kwatchgnupg makes knotifications and ktextwidgets optional~~

Drop unused dependency app-crypt/gpgme
commit 7dfe5e49299d02a7539047f8459c4c75ae803eec

### DIFF
--- a/kde-apps/kleopatra/kleopatra-9999.ebuild
+++ b/kde-apps/kleopatra/kleopatra-9999.ebuild
@@ -18,7 +18,7 @@ KEYWORDS=""
 IUSE=""
 
 # drop qtgui subslot operator when QT_MINIMAL >= 5.7.0
-COMMON_DEPEND="
+DEPEND="
 	$(add_frameworks_dep kcmutils)
 	$(add_frameworks_dep kcodecs)
 	$(add_frameworks_dep kconfig)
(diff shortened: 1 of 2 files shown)


kde-apps/kleopatra/kleopatra-9999.ebuild
```diff
@@ -15,10 +15,10 @@ HOMEPAGE="https://www.kde.org/applications/utilities/kleopatra"
 LICENSE="GPL-2+ handbook? ( FDL-1.2+ )"
 KEYWORDS=""
 
-IUSE=""
+IUSE="+kwatchgnupg"
```
johu: Is it really worth it to make it optional?

a17r: well since upstream considered the knotifications dep as 'heavy'... admittedly not for regular Linux desktop use. I couldn't yet check what it would change wrt total/implicit dependences, with all other Frameworks already pulled in considered.

a17r:
Optional kwatchgnupg indeed saves nothing, removed it again.

Due to the fixed dependency in ``kde-apps/libkleo`` it would even be possible to move it to tree ahead of KDE PIM (including only ``kpimtextedit`` and ``kmime``).

johu:
Thanks, picked.

PR #727, opened 2016-07-16
kde-apps/kdepim-addons: Drop use of KDEPIM_FEATURES

a17r:


johu:
Picked and squashed into 4df748fcad88f687edd136af09d27610fd611db6

PR #729, opened 2016-07-22
sys-libs/kpmcore: Add KAuth dependency.

stikonas:
Add KAuth dependency.


kensington:
Thanks, picked as 1951218458c65353d7d40d6ea3d0fad1e3661bae.

PR #734, opened 2016-07-26
kde-apps/akonadi-contact -> kde-apps/akonadi-contacts

a17r:
I've just seen digikam-5.0.0 is in tree already, that should get its akonadi-contact dep fixed as well (edit ebuild is enough?)


kensington:
Thanks!

I've fixed the digikam dependency in gentoo/gentoo@56d84d84bfa536eefb24edcd604d7b04a121c5cf. In theory inline should be fine because it's behind a masked USE flag and akonadi-contacts has only ever been available in the overlay.

PR #735, opened 2016-07-26
net-p2p/ktorrent: Add back USE=kross,rss,scanfolder

a17r:


kensington:
Thanks! Looks like they're almost back to feature parity. :-) Picked as 084c9732d53fa2910ec60a5ddd03608e33fab5b8.

PR #736, opened 2016-07-27
kde-apps/minuet: Add dependency on media-sound/fluidsynth

eliasp:

Build will fail otherwise:
```
-- Checking for module 'drumstick-alsa>=1.0.1'
--   Found drumstick-alsa, version 1.0.2
-- Checking for module 'drumstick-file>=1.0.1'
--   Found drumstick-file, version 1.0.2
-- Checking for module 'fluidsynth>=1.1.6'
--   
CMake Error at /usr/share/cmake/Modules/FindPkgConfig.cmake:424 (message):
  A required package was not found
Call Stack (most recent call first):
  /usr/share/cmake/Modules/FindPkgConfig.cmake:597 (_pkg_check_modules_internal)
  src/plugins/CMakeLists.txt:37 (pkg_check_modules
```
I don't know whether this is also required for older releases.

### DIFF
--- a/kde-apps/minuet/minuet-9999.ebuild
+++ b/kde-apps/minuet/minuet-9999.ebuild
@@ -27,6 +27,7 @@ COMMON_DEPEND="
 	$(add_qt_dep qtwidgets)
 	media-libs/alsa-lib
 	>=media-sound/drumstick-1.0.1
+	media-sound/fluidsynth
 "
 DEPEND="${COMMON_DEPEND}
 	virtual/pkgconfig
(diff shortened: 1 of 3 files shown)


kde-apps/minuet/minuet-9999.ebuild
```diff
@@ -27,6 +27,7 @@ COMMON_DEPEND="
 	$(add_qt_dep qtwidgets)
 	media-libs/alsa-lib
 	>=media-sound/drumstick-1.0.1
+	>=media-sound/fluidsynth-1.1.6
```
hmmh-johannes-huber: Minimum version operator is not needed, there is only 1.1.6 in gentoo main tree and the package was not bumped since a year...

 Please check cmake files on stable branch if it is needed for 16.08.

eliasp:
Verified + fixed…

johu:
Thanks, picked d0be523997acf6f44f24b4c0c1234ad062dc484d.

PR #738, opened 2016-07-31
media-gfx/digikam: Add USE=opencv3, translations and doc

a17r:
Built fine.

### DIFF
--- a/eclass/kde5.eclass
+++ b/eclass/kde5.eclass
@@ -522,6 +522,7 @@ EOF
 	# when required
 	if [[ -d po && -v LINGUAS ]] ; then
 		pushd po > /dev/null || die
+		local lang
 		for lang in *; do
 			if [[ -d ${lang} ]] && ! has ${lang} ${LINGUAS} ; then
 				rm -r ${lang} || die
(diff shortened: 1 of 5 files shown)


eclass/kde5-functions.eclass
```diff
@@ -300,6 +300,23 @@ kde_l10n2lingua() {
 	done
 }
 
+# @FUNCTION: prune_doc_linguas
+# @USAGE: <docdir>
+# @DESCRIPTION:
+# Removes all but wanted LINGUAS from specified doc dir.
+# No action if LINGUAS is unspecified.
+prune_doc_linguas() {
```
Arfrever: Add `local lang` here.

a17r:
see also: https://github.com/gentoo/gentoo/pull/2009

eclass/kde5.eclass
```diff
@@ -534,17 +535,20 @@ EOF
 				fi
 			fi
 		done
+		unset lang
```
Arfrever: When `local` is used for a variable, then previous state of this variable (e.g. it being unset or its previous value if it was set) will be automatically restored after finishing of execution of given function.

eclass/kde5.eclass
```diff
@@ -534,17 +535,20 @@ EOF
...
+	if [[ ${KDE_BUILD_TYPE} = release && ${CATEGORY} != kde-apps ]] ; then
+		if [[ ${KDE_HANDBOOK} != false && -d ${KDE_DOC_DIR} && -v LINGUAS ]] ; then
 			pushd ${KDE_DOC_DIR} > /dev/null || die
+			local lang
 			for lang in *; do
 				if ! has ${lang} ${LINGUAS} ; then
 					cmake_comment_add_subdirectory ${lang}
 				fi
 			done
+			unset lang
```
Arfrever: Same here.

Arfrever:
Maybe this USE flag is not needed at all?. You can use:
`-DENABLE_OPENCV3=$(has_version ">=media-libs/opencv-3" && echo yes || echo no)`

And `media-libs/opencv:=` in dependencies.

kensington:
Radical! Picked as 95cd09d47a77be8667dbf1d535a86ffdc151ffdf...774151db1c8bf23e3e0fbc65b418d60c48f775c5.

PR #739, opened 2016-08-01
kde-apps/kde4-l10n: Fix latest file collision

a17r:
Gentoo-bug: 589836


johu:
Thanks. Picked 89dfb034bcbf258011443982e04b22809307a25a.

PR #742, opened 2016-08-03
kde-apps/konqueror: Add block against kde-apps/kfmclient:4

a17r:


kensington:
Thanks, picked as 95cd09d47a77be8667dbf1d535a86ffdc151ffdf.

PR #756, opened 2016-08-14
net-libs/libkfbapi: Remove, dead API, dead upstream

a17r:
http://martys.typepad.com/blog/2015/03/facebook-chat-in-kde-telepathy-is-in-death-bed.html


kensington:
Thanks!

PR #757, opened 2016-08-14
kde-apps/konsole: Set KDE_HANDBOOK to optional, fix DEPENDs

a17r:
Ported away from kdelibs4support in upstream commit
27dec8d02f705c77eef45a2533bed203eec9454f


kensington:
Thanks!

PR #760, opened 2016-08-15
media-gfx/digikam: Add USE=jpeg2k,opengl,openmp,panorama and sort DEPENDs

a17r:
Also set some default USE flags.


kensington:
Thanks, picked as 97cb59df6ff1000fe834c5b1dd26d61b3630fd40.

PR #761, opened 2016-08-16
kde-apps/marble: Add USE=dbus,positioning fix DEPENDs, DESCRIPTION

a17r:
- Add Qt5SerialPort to aprs DEPENDs now that it's in tree
- Add Qt5{Location,Positioning} support, same USE naming as in PyQt5
- Other packages are using 'designer' so bring marble in line
- Disable marble-qt if USE=kde, it only adds confusion
- Drop unused and deprecated Qt5OpenGL
- Improve DESCRIPTION


kensington:
Thanks!

PR #763, opened 2016-08-19
kde-apps/ktp-l10n: Add blocker on <kde-apps/kde-l10n-16.08.0

MJDSys:
There are a number of files that changed packages between 16.04 and 16.08.
Add a blocker to get the update to work correctly.

### DIFF
--- a/app-cdr/k3b/k3b-9999.ebuild
+++ b/app-cdr/k3b/k3b-9999.ebuild
@@ -8,7 +8,7 @@ KDE_HANDBOOK="forceoptional"
 KDE_TEST="true"
 inherit kde5
 
-DESCRIPTION="Full-featured CD/DVD/Blu-ray burning and ripping application based on KDE Frameworks"
+DESCRIPTION="Full-featured burning and ripping application based on KDE Frameworks"
 HOMEPAGE="http://www.k3b.org/"
 
 LICENSE="GPL-2 FDL-1.2"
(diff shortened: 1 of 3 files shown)


a17r:
What blockers are there?

MJDSys:
I don't have the list anymore, but it had to do with the sr files.  kde-l10n-16.04 had some ktp related translation files, and after adding this change both have installed successfully.

If you want, I can try to reproduce and submit the list.

a17r:
This seems indeed related to L10N=sr splitting that was introduced with 16.04.3. The correct blocker should be <16.04.3 then, and also be added to kdepim-l10n, and may better be backported to tree as well.

MJDSys:
So it appears you are correct, trying to go from 16.04.2 to 16.04.3 gives the same result (see https://gist.github.com/MJDSys/57b986d345be07858b200867131b1174 for details).  kdepim-l10n-16.04.2 won't even install due to conflicts.

I think the simple answer is that 16.04.2's l10n has some problems, and the easiest thing to do is add conflicts for both ktp and kdepim to both 16.04.3 and 16.08.0.  That way anyone upgrading will get the right solution.  I'll update this PR, and if you are happy submit one against gentoo/gentoo as well.

a17r:
Thanks for fixing this! PR looks good except for that one bogus k3b commit that sneaked into it ;)

MJDSys:
@a17r That is from the kde repo hosted by gentoo.  It apparently hasn't been synced to github yet.

I'll submit a PR against main for 16.04.3.

kensington:
Thanks, picked as 076dffa869ba6536e223556755ebddf0e631d394...36e8a8179e1036075d1d73f175a0284e14c8d396.

PR #764, opened 2016-08-19
More KIO slaves

a17r:


kensington:
Should kio-gdrive be in kde-misc too?

a17r:
Hm, if the 'networking' in net-misc really only means 'local', then probably yes.

kensington:
Thanks, picked as e70a018f82ffdaa62c11bb582c4367b99cc07e38...534fa63f20ca055afb5c9aa43f64fe020d08aec0.

PR #776, opened 2016-08-27
kde-apps/ktp-l10n: Guard against cmake error w/ empty dir, cleanup L10N

a17r:
Gentoo-bug: 592266


kensington:
Thanks, picked as 73b18a5ab682a916efaee9b3fb075cd47439ddec.

PR #778, opened 2016-08-28
profiles: kdepim-16.08.0: Lift mask on kleopatra and deps

a17r:
It has use cases outside of PIM. None of the dependencies have
a blocker against OLD PIM, save for kleopatra itself.

Gentoo-bug: 592282


kensington:
Thanks, picked as e40c505e46a5a34c0be38c900207fc19357d62d3.

PR #781, opened 2016-08-31
dev-util/kdevelop-python: Switch to python-single-r1.eclass

a17r:


kensington:
Thanks, picked as cace01536c9fe6b10164993eae63b4b6ae36787d.

PR #783, opened 2016-09-01
sets + Documentation: Use 5.9999 instead of 9999 for kde-apps/kgpg

eliasp:


a17r:
I see where you're coming from, the fix for broken/conflicting sets would not be complete without removing `kde-apps/kuser` from kdeadmin-live, but a similar proposal was turned down in the past...

a17r:
obsolete by kgpg-9999 finally using KF5 :)

PR #793, opened 2016-09-15
kde-apps/calendarjanitor: Fix file collision

a17r:
console.categories currently only has an entry for konsolekalendar.

Gentoo-bug: 593886


kensington:
Thanks, picked as fb6905408ba20140176d73f1612d6fd8ec856596.

PR #799, opened 2016-10-08
kde-apps/messagelib: require at least Qt 5.7.0

a17r:
Added kdepim-16.08 package.unmask including Qt-5.7.0.
Removed kde-frameworks-5.25 set and Documentation.
Synced kde-frameworks/networkmanager-qt with tree.


kensington:
Thanks!

PR #800, opened 2016-10-08
New releases: choqok, kdb, kproperty, kreport

a17r:
kdb, kproperty, kreport pkgmoved to dev-db/ and dev-libs/


kensington:
Thanks!

PR #803, opened 2016-10-09
kde-apps: 16.08.2 version bump

a17r:
Doing the switch in here so there is a proper template for the upcoming 16.08.2 version bump.

(don't merge before 13th) ;)


kensington:
Thanks, picked as 942613f8cfd8e134421c39b0347a3387f7a2b5df...fde1631d7214893243ccb4501adc6193077327df.

PR #816, opened 2016-10-30
kde-frameworks/kwayland: qtgui dependency needs USE="egl"

eliasp:
Otherwise starting any kind of Plasma Wayland session (full, nested)
will fail with the error message:

`Failed to load client buffer integration: wayland-egl`

### DIFF
--- a/kde-frameworks/kwayland/kwayland-9999.ebuild
+++ b/kde-frameworks/kwayland/kwayland-9999.ebuild
@@ -16,12 +16,12 @@ IUSE=""
 
 DEPEND="
 	$(add_qt_dep qtconcurrent)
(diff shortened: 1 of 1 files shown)


johu:
Does dev-qt/qtwayland needs egl too?

eliasp:
> Does dev-qt/qtwayland needs egl too?

You're right - completely missed this & updated the PR. Thanks for pointing it out.

johu:
Thanks. Picked with commit message "kde-frameworks/kwayland: DEPEND on dev-qt/{qtgui,qtwayland}[egl]" 47be3ea7b6041ba6b2b77c8399cda31d59e4ee8c

PR #817, opened 2016-11-03
Eclass changes/fixes

a17r:
FRAMEWORKS_MINIMAL was always set by kde5-functions.eclass, making the assignments as done with KDE_AUTODEPS since 040cef12f24e64267d5c60bd0b2fbcc5a3280cec a no-op. Since that re-introduces 5.28.0 minimum for KDE Applications 16.12, adding also KDEBASE=kdepim to be able to fix deps for PIM only.

This PR moves assignment of the various KDE related *_MINIMAL variables from ``kde5-functions.eclass`` to ``kde5.eclass`` if ``KDE_AUTODEPS=true``. Effectively no change for the vast majority of ebuilds using kde5.eclass.

### DIFF
--- a/eclass/kde5-functions.eclass
+++ b/eclass/kde5-functions.eclass
@@ -23,6 +23,35 @@ case ${EAPI} in
 	*) die "EAPI=${EAPI:-0} is not supported" ;;
 esac
 
+# determine the build type
+if [[ ${PV} = *9999* ]]; then
+	KDE_BUILD_TYPE="live"
+else
+	KDE_BUILD_TYPE="release"
+fi
+export KDE_BUILD_TYPE
+
+case ${CATEGORY} in
+	kde-frameworks)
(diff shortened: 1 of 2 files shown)


johu:
👎  The minimum declarations unset sounds a little bit strange. In kde4*eclass we had a override mechanism. Maybe this is a better aproach.

a17r:
I kind of like my solution ;) Minimum declarations now happen in one place, in one eclass instead of two, and everything can be overridden by ebuilds without adding another variable (3 times). Someone who only uses kde5-functions.eclass will have to set FRAMEWORKS_MINIMAL manually, if that is even necessary for that package. What good are those min versions for non-KDE_AUTODEPS anyway?

a17r:
Pushed new approach. If you like that, now that I look at it, one could also move QT_MINIMAL to kde5.eclass like the others, should it ever become necessary to require this for e.g. all of a certain kde-plasma version.

a17r:
Reverse approach, moving *_MINIMAL declarations into kde5-functions.eclass

kensington:
Thanks!

PR #822, opened 2016-11-06
media-gfx/digikam: Add subslot op, improve openmp check

a17r:
Thanks to Soap.


johu:
Thanks, picked.

PR #828, opened 2016-11-17
[WIP] kde-apps/kopete: Add kf5 version

a17r:
Just to put it out there - the ebuild has existed for some time, but I can't say it is doing much besides segfault at startup.

However, someone upstream just called it 'near complete' so maybe I'm just missing something.


johu:
Please decide yourself if it is needed.

PR #833, opened 2016-11-25
kde-apps/gwenview: Drop unused cmake variable

a17r:
Beware of untested crap(TM) for I'm on battery with the laptop until the evening.


johu:
We dont backport features if they are not critical....

a17r:
Aw would have been nice though - it's a feature that is often asked for. And I expect no breakage of the patch, that is just making use of an otherwise dead source directory and installing a separate binary (no effect on the main application).

johu:
https://wiki.gentoo.org/wiki/Project:KDE/Coding_style#Patches

kensington:
Thanks, picked as 04292b031e4e7bbf18758cf83df9123bd278119d.

I added some upstream metadata that I found to the commit message for posterity, hope that's OK.

PR #837, opened 2016-11-27
kde-apps/kdecore-meta: Re-add ported baseapps

a17r:
Historically, these were part of kdebase-meta.


kensington:
Thanks!

PR #850, opened 2016-12-20
Add kdecore-meta and kdeutils-meta use flags

foofaafoo:
kdeutils-meta: add pim use flag
kdecore-meta: add webengine use flag


johu:
Sorry but this makes no sense to add use flags to a meta package, where meta packages are designed to pull in all possible dependencies with some corner case exceptions. Portage provides a lot of possibilities to create your own meta package/set to create such use case.

PR #851, opened 2016-12-20
Sync with tree

a17r:
See also:

https://github.com/gentoo/gentoo/pull/3189
https://github.com/gentoo/gentoo/pull/3190
https://github.com/gentoo/gentoo/pull/3191


kensington:
Thanks!

PR #852, opened 2016-12-21
kde-apps/okular: Raise DEPEND to >=app-text/poppler-0.50.0

a17r:
okular already ifdefs support for latest poppler version, rather than
add another subslot operator, just raise the minimum version.

Upstream commit 2868e3e5ca802dd4decfdeb1dded82394a4c34b6

### DIFF
--- a/kde-apps/okular/okular-9999.ebuild
+++ b/kde-apps/okular/okular-9999.ebuild
@@ -47,7 +47,7 @@ DEPEND="
 		virtual/jpeg:0
 	)
 	mobi? ( $(add_kdeapps_dep kdegraphics-mobipocket) )
-	pdf? ( app-text/poppler[qt5,-exceptions(-)] )
+	pdf? ( >=app-text/poppler-0.50.0[qt5,-exceptions(-)] )
 	postscript? ( app-text/libspectre )
 	speech? ( $(add_qt_dep qtspeech) )
 	tiff? ( media-libs/tiff:0 )


johu:
50aa7d41d6258caefdbd95a7c5e08a645fd1f501 allows to compile with lower versions.

a17r:
Sure, but so far okular didn't seem to need a subslot operator - to me it would be fine for the live ebuild to depend on latest tree version of poppler (any ~arch user will have it installed already), instead of triggering possibly unnecessary rebuilds in the future.

johu:
This is already possible without increase the version depdendency.

kensington:
> This is already possible without increasing the version depdendency.

how?

johu:
okular doesn't need to use a subslot from poppler, so why force user to higher version?

kensington:
Right, okular traditionally has not needed a subslot operator on poppler because it links to poppler-qt4, but consider kde/okular@2868e3e5ca802dd4decfdeb1dded82394a4c34b6:

```
#ifdef HAVE_POPPLER_0_50
    const Poppler::LinkOCGState *popplerLink = action->nativeId().value<const Poppler::LinkOCGState *>();
    pdfdoc->optionalContentModel()->applyLink( const_cast< Poppler::LinkOCGState* >( popplerLink ) );
#else
    (void)action;
#endif
}
```

If we neither raise the dependency nor add a subslot operator, what will happen when the user upgrades 0.49.0 -> 0.50.0?

PR #854, opened 2017-02-03
games-board/knights: fix LICENSE, add SELinux module, add USE=speech

awilfox:
* License is GPL-2+ as noted [in upstream source files](https://cgit.kde.org/knights.git/tree/src/board.cpp) and its [AppData XML manifest](https://cgit.kde.org/knights.git/tree/src/org.kde.knights.appdata.xml#n5).

* Package was missing KDE_SELINUX_MODULE="games" as other KDE Games packages have

* Upstream supports QtSpeech for speaking opponent moves

Package-Manager: portage-2.3.1


johu:
Please shorten commit summary or make individual commits.

johu:
Thanks.

PR #855, opened 2017-04-12
net-p2p/ktorrent: Add back USE=search

stikonas:
Upstream commit acc03acf1a1b9aa459a05eee513dda0de8d2c724


net-p2p/ktorrent/ktorrent-9999.ebuild
```diff
@@ -69,6 +69,10 @@ COMMON_DEPEND="
 		$(add_frameworks_dep kdewebkit)
 		$(add_kdeapps_dep syndication)
 	)
+	search? (
+		$(add_frameworks_dep kdewebkit)
+		dev-qt/qtwebkit:5
```
a17r: Please use `$(add_qt_dep ...)` as seen with the other Qt deps.

a17r:
Thanks!

PR #858, opened 2017-06-29
net-p2p/ktorrent: update deps.

stikonas:


net-p2p/ktorrent/ktorrent-9999.ebuild
```diff
@@ -57,24 +54,36 @@ COMMON_DEPEND="
...
 		$(add_frameworks_dep kdewebkit)
 		$(add_qt_dep qtwebkit)
 	)
-	shutdown? ( $(add_plasma_dep plasma-workspace) )
+	shutdown? (
+		$(add_frameworks_dep kdelibs4support)
+		$(add_plasma_dep plasma-workspace)
+	)
 	stats? ( $(add_frameworks_dep kplotting) )
+	upnp? (
```
a17r: this can be done on one line, like the one above

net-p2p/ktorrent/ktorrent-9999.ebuild
```diff
@@ -56,25 +54,35 @@ COMMON_DEPEND="
 	$(add_qt_dep qtdbus)
 	$(add_qt_dep qtgui)
 	$(add_qt_dep qtnetwork)
+	$(add_qt_dep qttest)
```
a17r: having set KDE_TEST=optional way above means that you should drop this line here, eclass will do

PR #860, opened 2017-08-20
media-gfx/ktikz: add ebuild

jfmcarreira:
Suggestion for the Ktikz app ebuild


a17r:
I'm sorry for letting this sit for so long. Is this project still active upstream - I'm looking at only a handful commits since end of last year. If so, and you are still willing to work on the ebuild as a maintainer, please instead of kde overlay you can submit it in one of two places:

https://wiki.gentoo.org/wiki/Project:GURU
https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers

PR #862, opened 2017-10-04
Update falkon-9999.ebuild

perfect7gentleman:


a17r:
Thanks, picked this PR while fixing the Manifest in place.

- Use git-amend and force-push if you need to fix an existing PR, here you ended up with two broken ones
- I fixed your git summary according to Gentoo's git workflow guidelines, please have a look.

PR #864, opened 2017-10-10
www-client/falkon: Fix logic for disabling D-Bus

jktjkt:
...because it was actually implemented backwards.

### DIFF
--- a/www-client/falkon/falkon-9999.ebuild
+++ b/www-client/falkon/falkon-9999.ebuild
@@ -63,7 +63,7 @@ src_prepare() {
 
 src_configure() {
 	local mycmakeargs=(
-		-DDISABLE_DBUS=$(usex dbus)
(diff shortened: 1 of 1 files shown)


www-client/falkon/falkon-9999.ebuild
```diff
@@ -63,7 +63,7 @@ src_prepare() {
 
 src_configure() {
 	local mycmakeargs=(
-		-DDISABLE_DBUS=$(usex dbus)
+		-DDISABLE_DBUS=$(usex dbus no yes)
```
a17r: thanks, just use `usex !dbus` instead ;)

jktjkt: OK, I didn't know that. Feel free to redo that bit and merge as you wish, I don't have any real interest in that package.

PR #865, opened 2018-07-08
CMake EAPI7 updates

SoapGentoo:
Let's first update the overlay before we move this to the main tree

### DIFF
--- a/eclass/cmake-utils.eclass
+++ b/eclass/cmake-utils.eclass
@@ -50,10 +50,12 @@ _CMAKE_UTILS_ECLASS=1
 # Set to enable in-source build.
 
 # @ECLASS-VARIABLE: CMAKE_MAKEFILE_GENERATOR
+# @DEFAULT_UNSET
 # @DESCRIPTION:
 # Specify a makefile generator to be used by cmake.
 # At this point only "emake" and "ninja" are supported.
-: ${CMAKE_MAKEFILE_GENERATOR:=emake}
+# In EAPI 7 and above, the default is set to "ninja",
+# whereas in EAPIs below 7, it is set to "emake".
 
 # @ECLASS-VARIABLE: CMAKE_MIN_VERSION
 # @DESCRIPTION:
@@ -112,7 +114,13 @@ esac
 inherit toolchain-funcs ninja-utils flag-o-matic multiprocessing xdg-utils
 
 case ${EAPI} in
-	5|6) inherit eutils multilib ;;
+	[56])
+		: ${CMAKE_MAKEFILE_GENERATOR:=emake}
+		inherit eapi7-ver eutils multilib
+		;;
+	*)
+		: ${CMAKE_MAKEFILE_GENERATOR:=ninja}
+		;;
 esac
 
 EXPORT_FUNCTIONS src_prepare src_configure src_compile src_test src_install
--- a/eclass/cmake-utils.eclass
+++ b/eclass/cmake-utils.eclass
@@ -116,16 +116,18 @@ inherit toolchain-funcs ninja-utils flag-o-matic multiprocessing xdg-utils
 case ${EAPI} in
 	[56])
 		: ${CMAKE_MAKEFILE_GENERATOR:=emake}
+		_cmake_eqawarn_command=eqawarn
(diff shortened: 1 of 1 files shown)


eclass/cmake-utils.eclass
```diff
@@ -53,7 +53,8 @@ _CMAKE_UTILS_ECLASS=1
 # @DESCRIPTION:
 # Specify a makefile generator to be used by cmake.
 # At this point only "emake" and "ninja" are supported.
-: ${CMAKE_MAKEFILE_GENERATOR:=emake}
```
mgorny: Wouldn't this introduce eclassdoc warning about no value?

eclass/cmake-utils.eclass
```diff
@@ -507,7 +507,8 @@ cmake-utils_src_configure() {
 	debug-print-function ${FUNCNAME} "$@"
 
 	if [[ ! ${_CMAKE_UTILS_SRC_PREPARE_HAS_RUN} ]]; then
-		eqawarn "cmake-utils_src_prepare has not been run, please open a bug on https://bugs.gentoo.org/"
+		local msg="cmake-utils_src_prepare has not been run, please open a bug on https://bugs.gentoo.org/"
+		[[ ${EAPI} != [56] ]] && die "${msg}" || eqawarn "${msg}"
```
mgorny: Maybe you want to make other eqawarn uses fatal as well.

kensington: Fine with me

kensington:
As this is such a widely used eclass, please send to -dev for review.

I'm concerned about switching the default to ninja as this is known to cause breakage in various packages, but if you believe the performance gain is worth it and -dev agrees I won't oppose that change.

junghans:
I just saw a bunch of ebuilds in the tree using `EAPI=7` already, but you are right, no tinderbox run needed yet. However, besides the performance question, this will break the packages with Fortran in them, so `fortran-2.eclass` should set `CMAKE_MAKEFILE_GENERATOR="emake"` (or we need to package [Kitware's ninja-fortran](https://github.com/Kitware/ninja) and support).

SoapGentoo:
Blocking non-recursive builds due to 10 packages using FORTRAN with CMake is inconsiderate to those users with 16-core machines. You can always add the `CMAKE_MAKEFILE_GENERATOR` to those ebuilds that require it.

eclass/cmake-utils.eclass
```diff
@@ -116,16 +116,18 @@ inherit toolchain-funcs ninja-utils flag-o-matic multiprocessing xdg-utils
 case ${EAPI} in
 	[56])
 		: ${CMAKE_MAKEFILE_GENERATOR:=emake}
+		_cmake_eqawarn_command=eqawarn
 		inherit eapi7-ver eutils multilib
 		;;
 	*)
 		: ${CMAKE_MAKEFILE_GENERATOR:=ninja}
+		_cmake_eqawarn_command=die
```
mgorny: To be honest, this wasn't exactly my intent. I meant making existing QA warnings fatal in EAPI 7 but starting with eqawarn for future warnings ;-).

eclass/cmake-utils.eclass
```diff
@@ -208,8 +210,8 @@ _cmake_check_build_dir() {
 		# Respect both the old variable and the new one, depending
 		# on which one was set by the ebuild.
 		if [[ ! ${BUILD_DIR} && ${CMAKE_BUILD_DIR} ]]; then
-			eqawarn "The CMAKE_BUILD_DIR variable has been renamed to BUILD_DIR."
-			eqawarn "Please migrate the ebuild to use the new one."
+			${_cmake_eqawarn_command} "The CMAKE_BUILD_DIR variable has been renamed to BUILD_DIR.
```
mgorny: This is a case for `eerror`+`die`.

eclass/cmake-utils.eclass
SoapGentoo: Please do it yourself then, this has cost me too much time already

a17r:
Merged 2/3 commits to overlay for testing, thanks, redid the third one myself.

PR #869, opened 2018-09-12
kde-apps/{kldap,libksieve}: fix use_if_iuse usage

mm1ke:
Hi,

Both packages (kldap, libksieve) using the function ```use_if_iuse``` from the ```eutils``` eclass but doesn't inherit it. This PR fixes that.

please review.


a17r:
Please port away from use_if_iuse rather than re-inheriting eutils.

a17r:
(I see no reason why this should not be `use` instead of `use_if_iuse`

mm1ke:
@a17r 
I've updated the PR and use ```use``` now instead of  ```use_if_iuse```
To be honest, i don't know why ```use_if_iuse``` was used here, but I also don't see a reason why not just use ```use```

PR #875, opened 2019-03-13
*/*: fix redirection

mm1ke:
Hi,

For some time now ```www.kde.org``` redirects to ```kde.org```. This PR updates all affected ebuilds to reflect this change. However, i know this is not a real problem as it doesn't really fix anything, which is why i also made a single commit for the changes. I'm ok if this PR won't get merged, but it would be nice if we could fix this :)

Please review.

### DIFF
--- a/app-editors/okteta/okteta-9999.ebuild
+++ b/app-editors/okteta/okteta-9999.ebuild
@@ -14,7 +14,7 @@ if [[ ${KDE_BUILD_TYPE} = release ]]; then
 fi
 
 DESCRIPTION="Hex editor by KDE"
-HOMEPAGE="https://www.kde.org/applications/utilities/okteta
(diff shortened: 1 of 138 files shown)


mm1ke:
@a17r 
Updated, i already had the eclass included but i put it in a separate commit now. :)

a17r:
> @a17r
> Updated, i already had the eclass included but i put it in a separate commit now. :)

Thanks, correct prefix for eclasses is: `kde5.eclass: `...

mm1ke:
> > @a17r
> > Updated, i already had the eclass included but i put it in a separate commit now. :)
> 
> Thanks, correct prefix for eclasses is: `kde5.eclass: `...

Fixed :)

a17r:
Thanks, please don't forget sign-off next time.

PR #879, opened 2019-05-02
KDE Partition Manager 4.0.0

stikonas:
Version bump request.

Added `kdbusaddons` dependency, removed `util-linux` in partitionmanager ebuild
Also added ~arm64 keywords.


a17r:
Thanks! Pushed 4.0.0 bumps directly to gentoo.git.

PR #882, opened 2019-11-10
kde-frameworks/kirigami: dev-qt/qtgraphicaleffects is a runtime depen…

stikonas:
…dency

Package-Manager: Portage-2.3.76, Repoman-2.3.16
Signed-off-by: Andrius Štikonas <andrius@stikonas.eu>


a17r:
thanks!

PR #884, opened 2019-11-16
cmake.eclass: New eclass

a17r:


eclass/cmake.eclass
```diff
@@ -54,30 +52,19 @@ _CMAKE_UTILS_ECLASS=1
...
+# The default is set to "ninja".
+: ${CMAKE_MAKEFILE_GENERATOR:=ninja}
 
 # @ECLASS-VARIABLE: CMAKE_REMOVE_MODULES_LIST
 # @DESCRIPTION:
-# Space-separated list of CMake modules that will be removed in $S during src_prepare,
-# in order to force packages to use the system version.
+# Space-separated list of CMake modules that will be removed in $S during
+# src_prepare, in order to force packages to use the system version.
 : ${CMAKE_REMOVE_MODULES_LIST:=FindBLAS FindLAPACK}
```
mgorny: I'd add a comment like 'Set to empty to disable removing modules entirely'.

mgorny: Also, maybe add array support here?

eclass/cmake.eclass
```diff
@@ -768,11 +583,11 @@ _cmake_emake_src_make() {
 
 }
 
-# @FUNCTION: cmake-utils_src_make
+# @FUNCTION: cmake_src_make
```
mgorny: To be honest, I always found the name `src_make` weird. Maybe `cmake_build`?

a17r: and then, likewise `_cmake_ninja_src_make`, `_cmake_emake_src_make` to `cmake_*_build`?

mgorny: I don't know. I don't think they need to be public API. In fact, I would personally just inline them in some little `case ... in`.

eclass/cmake.eclass
```diff
@@ -768,25 +587,33 @@ _cmake_emake_src_make() {
...
 # All arguments are passed to emake.
-cmake-utils_src_make() {
+cmake_build() {
 	debug-print-function ${FUNCNAME} "$@"
 
 	_cmake_check_build_dir
 	pushd "${BUILD_DIR}" > /dev/null || die
 
-	_cmake_${CMAKE_MAKEFILE_GENERATOR}_src_make "$@"
+	_cmake_${CMAKE_MAKEFILE_GENERATOR}_build "$@"
```
junghans: Could we use `cmake --build .` here?

eclass/cmake.eclass
```diff
@@ -819,15 +646,16 @@ cmake-utils_src_test() {
...
 # @DESCRIPTION:
 # Function for installing the package. Automatically detects the build type.
-cmake-utils_src_install() {
+cmake_src_install() {
 	debug-print-function ${FUNCNAME} "$@"
 
 	_cmake_check_build_dir
 	pushd "${BUILD_DIR}" > /dev/null || die
-	DESTDIR="${D}" ${CMAKE_MAKEFILE_GENERATOR} install "$@" || die "died running ${CMAKE_MAKEFILE_GENERATOR} install"
+	DESTDIR="${D}" ${CMAKE_MAKEFILE_GENERATOR} install "$@" ||
```
junghans: And `cmake --install` here.

eclass/cmake.eclass
a17r: What's the benefit on giving up our established wrappers?

PR #887, opened 2020-01-27
upcoming ppc64 dep keywording

gyakovlev:


gyakovlev:
qtcharts is now keyworded in ::gentoo

gyakovlev:
was merged, thanks!

PR #894, opened 2020-11-30
Fixed markdownpart not git cloning

XutaxKamay:
This is a temporary fix for https://invent.kde.org/kde/markdownpart.git
not redirecting to https://invent.kde.org/utilities/markdownpart.git
Instead we get prompted with login,
This itself might be needed to be reported at KDE.


a17r:
Thanks for noticing. This will be indeed not a temporary fix, at least for the upstream side, as the `kde` redirect only works for packages available at the time of the transition from cgit/phabricator to gitlab, not ones created afterwards. The `KDE_ORG_CATEGORY` var could only be dropped once we settle for `utilities` to be associated with the `kde-misc` category, but right now there is probably a mix of all kinds of upstream categories in there.


If you want me to merge merge your PR you should drop the superfluous empty line after `KDE_ORG_NAME` and change commit message (apart from updating per info gained here) according to Gentoo's git workflow guidelines: https://wiki.gentoo.org/wiki/Gentoo_git_workflow#Commit_message_format

Otherwise I'll do the fix and give credits via `Thanks-to` tag in git summary.

XutaxKamay:
I see, and yes like you said I don't think this is the only package that got this problem but I didn't know if it was already known by maintainers (you) of this repository, so I tried to point it out.

I'm new to Gentoo world, so thanks for pointing me that.
Tell me if the formatting looks correct, otherwise I guess you'll have to do it yourself, I'm sorry.

a17r:
> I think Suggested-by might be better I guess?

For now, while you haven't set yourself up with our git workflow yet, I just need your `Signed-off-by:` with `first name` `last name` `<e-mail>`

XutaxKamay:
Alright, thanks I'll do that.
I feel like a retard now, but I'll get better at this I promise.

a17r (approved):
Thanks! When you're new to Gentoo and already dive into contributing ebuild fixes then you are way ahead of myself at that time. ;)

PR #901, opened 2021-04-19
kde-plasma/plasma-workspace: Rename qalculate USE flag to calculator

zpuskas:


a17r:
Nice work! Maybe you can create the patch using `git format-patch` directly from your upstream MR so we have a proper patch header and reference to the MR. ;)

a17r (changes requested):
New suggestion:

In `src_prepare()`, just add the following line:
`use calculator || cmake_run_in runners cmake_comment_add_subdirectory calculator`

That works for both stable branch and live ebuild without using a patch, and we can drop the line whenever you got your patch upstreamed.

PR #902, opened 2021-07-15
kde-misc/skanpage: import from ::guru

AndrewAmmerlaan:
A scanning application similar to skanlite, but unlike skanlite it supports scanning to pdf and re-ordering the scanned pages.

No release yet, but it already works pretty well, there is also a snapshot ebuild for skanpage in ::guru

Package-Manager: Portage-3.0.20, Repoman-3.0.3
Signed-off-by: Andrew Ammerlaan <andrewammerlaan@gentoo.org>

### DIFF
--- a/media-gfx/skanpage/skanpage-22.04.49.9999.ebuild
+++ b/media-gfx/skanpage/skanpage-22.04.49.9999.ebuild
@@ -1,0 +1,41 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+KDE_ORG_CATEGORY="utilities"
+PVCUT=$(ver_cut 1-3)
+KFMIN=5.91
+QTMIN=5.15.2
+inherit ecm kde.org
+
+DESCRIPTION="Multi-page scanning application supporting image and pdf files"
+HOMEPAGE="https://apps.kde.org/skanpage/"
+
+LICENSE="|| ( GPL-2 GPL-3 ) CC0-1.0"
+SLOT="5"
+KEYWORDS=""
+IUSE=""
+
+DEPEND="
+	>=dev-qt/qtconcurrent-${QTMIN}:5
+	>=dev-qt/qtcore-${QTMIN}:5
+	>=dev-qt/qtdeclarative-${QTMIN}:5
+	>=dev-qt/qtgui-${QTMIN}:5
+	>=dev-qt/qtnetwork-${QTMIN}:5
+	>=dev-qt/qtprintsupport-${QTMIN}:5
+	>=dev-qt/qtquickcontrols2-${QTMIN}:5
+	>=dev-qt/qtwidgets-${QTMIN}:5
+	>=kde-apps/libksane-${PVCUT}:5
+	>=kde-frameworks/kconfig-${KFMIN}:5
+	>=kde-frameworks/kcoreaddons-${KFMIN}:5
+	>=kde-frameworks/kcrash-${KFMIN}:5
+	>=kde-frameworks/ki18n-${KFMIN}:5
+	>=kde-frameworks/kio-${KFMIN}:5
+	>=kde-frameworks/kirigami-${KFMIN}:5
+	>=kde-frameworks/kjobwidgets-${KFMIN}:5
+	>=kde-frameworks/kwidgetsaddons-${KFMIN}:5
+	>=kde-frameworks/kxmlgui-${KFMIN}:5
+	>=kde-frameworks/purpose-${KFMIN}:5
+"
+RDEPEND="${DEPEND}"
--- a/media-gfx/skanpage/metadata.xml
+++ b/media-gfx/skanpage/metadata.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="project">
+		<email>kde@gentoo.org</email>
+		<name>Gentoo KDE Project</name>
+	</maintainer>
+</pkgmetadata>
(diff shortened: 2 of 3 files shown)


AndrewAmmerlaan:
> No release yet, but it already works pretty well, there is also a snapshot ebuild for skanpage in ::guru

Skanpage has made a first release, ebuild updated accordingly.

kde-misc/skanpage/skanpage-9999.ebuild
```diff
@@ -0,0 +1,34 @@
...
+DESCRIPTION="Multi-page scanning application supporting image and pdf files"
+HOMEPAGE="https://invent.kde.org/utilities/skanpage"
+
+LICENSE="|| ( GPL-2 GPL-3 ) CC0-1.0"
+SLOT="5"
+
+DEPEND="
+	>=dev-qt/qtwidgets-${QTMIN}:5
+	>=dev-qt/qtprintsupport-${QTMIN}:5
+	>=dev-qt/qtquickcontrols2-${QTMIN}:5
```
a17r, line 24: alphanumerical sorting.

kde-misc/skanpage/skanpage-9999.ebuild
```diff
@@ -0,0 +1,34 @@
...
+	>=kde-apps/libksane-21.12.3:5
```
a17r, line 25: Depend on the same version of libksane as $PN within KDE Gear release group.

grep for `PVCUT` inside kde overlay for how to do it.

kde-misc/skanpage/skanpage-9999.ebuild
```diff
@@ -0,0 +1,34 @@
...
+	>=kde-frameworks/kcoreaddons-${KFMIN}:5
+	>=kde-frameworks/ki18n-${KFMIN}:5
+	>=kde-frameworks/kio-${KFMIN}:5
+	>=kde-frameworks/kirigami-${KFMIN}:5
+	>=kde-frameworks/kjobwidgets-${KFMIN}:5
+	>=kde-frameworks/kwidgetsaddons-${KFMIN}:5
+	>=kde-frameworks/kxmlgui-${KFMIN}:5
```
a17r, line 32: Are you sure these deps are complete? Did you check linking?

kde-misc/skanpage/skanpage-9999.ebuild
```diff
@@ -0,0 +1,34 @@
+# Copyright 1999-2021 Gentoo Authors
```
a17r, line 1: Update Copyright.

a17r (changes requested):
Thanks for your effort.

- 22.03.80 is not a release, it was a beta.
- Please create 22.04.49.9999 and 9999 ebuilds instead, release version will be for gentoo.git
- Are we sure `kde-misc` is the right choice here?

kde-misc/skanpage/skanpage-9999.ebuild
AndrewAmmerlaan: They are now. This package has had quite the rewrite since I first wrote this ebuild and it seems like it has gained a few extra dependencies along the way.

AndrewAmmerlaan:
> * Are we sure `kde-misc` is the right choice here?

No, but `kde-misc` is where `skanlite` is, and since skanpage and skanlite are very closely related applications I thought it would be good to have them in the same category (though perhaps `kde-apps` would be better for both).

a17r:
kde-apps is deprecated and to be abandoned long-term, kde-misc is not really a good fit (even if skanlite sets a bad precedent):

```
The kde-misc category contains various miscellaneous KDE packages.
In general, packages here are extensions of the Plasma desktop environment
(ioslaves, command center modules, plasmoids, ...), while KDE applications
that do not rely on the desktop environment better suit other categories.
```

AndrewAmmerlaan:
> kde-apps is deprecated and to be abandoned long-term, kde-misc is not really a good fit (even if skanlite sets a bad precedent):
> 
> ```
> The kde-misc category contains various miscellaneous KDE packages.
> In general, packages here are extensions of the Plasma desktop environment
> (ioslaves, command center modules, plasmoids, ...), while KDE applications
> that do not rely on the desktop environment better suit other categories.
> ```

Alright, so what would be a good fit for this package? `media-gfx` maybe? This is where sane and friends live and it is also where some scanner drivers are (e.g. brscan4). Should we also move skanlite there for consistency?

thesamesam:
Yeah, let's do media-gfx.

a17r:
Let's keep skanlite where it is right now, in case we come up with a better fitting category in the future.

thesamesam:
OK, np.

AndrewAmmerlaan:
> Yeah, let's do media-gfx. Discussed it on IRC too and we felt it's the best probably, in lieu of a better suggestion.

Alright, moved to `media-gfx/skanpage`.

media-gfx/skanpage/skanpage-22.04.49.9999.ebuild
```diff
@@ -0,0 +1,41 @@
...
+HOMEPAGE="https://apps.kde.org/skanpage/"
+
+LICENSE="|| ( GPL-2 GPL-3 ) CC0-1.0"
+SLOT="5"
+KEYWORDS=""
+IUSE=""
+
+DEPEND="
+	>=dev-qt/qtconcurrent-${QTMIN}:5
+	>=dev-qt/qtcore-${QTMIN}:5
```
a17r, line 22: `dev-qt/qtcore` is always provided by `ecm.eclass`, you can drop it.

a17r (changes requested):
lgtm, provided ependencies

media-gfx/skanpage/skanpage-22.04.49.9999.ebuild
thesamesam: I'll fix on merge.

thesamesam:
Thank you!

media-gfx/skanpage/skanpage-22.04.49.9999.ebuild
AndrewAmmerlaan: > I'll fix on merge.

Thanks

PR #905, opened 2022-02-15
dev-util/arcanist: remove dep on deprecated libphutil

telans:
treecleaned deprecated libphutil. see https://github.com/phacility/libphutil


telans:
A side note, would it be possible to get this into ::gentoo with a snapshot?

a17r:
This package is going to be removed from kde overlay. We have no use for it any longer.

https://www.mail-archive.com/gentoo-dev@lists.gentoo.org/msg89422.html

PR #907, opened 2022-06-21
ecm.eclass: Workaround Portage depgraph shortcomings from revdeps' side

a17r:
Bug: https://bugs.gentoo.org/836726

I hate having to do that, but the alternative is to disable the option entirely.


Pesa (approved):
I left a [comment](https://bugs.gentoo.org/836726#c27) on bugzilla, but the choice is really up to you.

PR #909, opened 2022-07-19
app-cdr/dolphin-plugins-mountiso: fix LICENSE

mm1ke:
Hi @a17r 

I saw ther're a lot of ebuilds were the `LICENSE` still needs to be validated. If you don' t mind i can have a look at them and fix them like i did now with `app-cdr/dolphin-plugins-mountiso`
I usually check a few upstream source files to find out the `LICENSE`, anything else i should consider?
In this case it was pretty simple, since the source files state:
```
SPDX-License-Identifier: GPL-2.0-or-later
```

Signed-off-by: Michael Mair-Keimberger <mmk@levelnine.at>
Portage 3.0.32 / pkgdev 0.2.1 / pkgcheck 0.10.11

### DIFF
--- a/app-cdr/dolphin-plugins-mountiso/dolphin-plugins-mountiso-22.07.80.ebuild
+++ b/app-cdr/dolphin-plugins-mountiso/dolphin-plugins-mountiso-22.07.80.ebuild
@@ -15,7 +15,7 @@ inherit ecm gear.kde.org
 DESCRIPTION="Dolphin plugin for ISO loopback device mounting"
 HOMEPAGE="https://apps.kde.org/dolphin_plugins/"
 
-LICENSE="GPL-2" # TODO: CHECK
+LICENSE="GPL-2+"
 SLOT="5"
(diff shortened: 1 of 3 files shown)


a17r:
Thanks, much appreciated! This comes from pre kde.org.eclass/ecm.eclass era setting default LICENSE. I'm sure there should be some kind of tool to scan out SPDX headers. Also, there would be separate handbook IUSE LICENSEs to consider, for which there are existing examples.

mm1ke:
You're welcome. :)
I had a quick look at some tools which are available in gentoo but couldn't find anything, except some which check licenses for go packages. I guess for now i'll stick with simply checking them ;)
Regarding handbook LICENSES, i guess you mean something like in `skanlite`: `|| ( GPL-2 GPL-3 ) handbook? ( FDL-1.2+ )`?
In that case, i'll look out for it, cerntainly wouldn't have checked them in the first place.

PR #911, opened 2022-09-16
kde-plasma/kpipewire: use plasma.kde.org eclass

xarblu:
kde-plasma/kpipewire-5.25.90 doesn't fetch anything with the regular `kde.org` eclass and throws an error that "The source directory '${S}' doesn't exist". Also changed the eclass for the other ebuilds for consistency.


a17r (approved):
Thanks!

PR #913, opened 2022-11-12
*/*: HOMEPAGE fixes

mm1ke:
Hi,

Just some more minor `HOMEPAGE` fixes.


a17r:
It is a bit odd that they are `www`ing again, will try to find out if that is an intended (permanent) change.

a17r:
Thanks a lot!

PR #914, opened 2022-11-19
dev-util/kdevelop-python: upstream added py3.10

t0b3:


t0b3:
@a17r this will add py310

dev-util/kdevelop-python/kdevelop-python-22.12.49.9999.ebuild
```diff
@@ -6,7 +6,7 @@ EAPI=8
 ECM_TEST="forceoptional"
 KDE_ORG_CATEGORY="kdevelop"
 KDE_ORG_NAME="kdev-python"
-PYTHON_COMPAT=( python3_{8..9} )
+PYTHON_COMPAT=( python3_{8..10} )
```
AndrewAmmerlaan, line 9: Could you also check for `python3_11` compatibility? It will save us time later if we can add both targets in one go.

t0b3: I considered this for a second but realized upstream doesn't support it yet...

see https://github.com/KDE/kdev-python/blob/master/CMakeLists.txt#L46

... or would you like to go further than upstream?

AndrewAmmerlaan: We could patch the cmake file to allow for `python3_11`, but we can only do this if it will actually work with `python3_11` (i.e. if all tests pass). The ebuild seems to RESTRICT execution of the tests so unless we can fix that we should probably just stick to whatever upstream lists as supported.

a17r:
Thanks!

a17r (approved):

PR #916, opened 2023-01-19
dynlink-scanner fixes

dlan17:
two minor fixes for dynlink-scanner

dynlink-scanner: drop the obsolete '-e' option of qfile
dynlink-scanner: add pie-executable to emime type detection list


a17r (approved):
I wasn't aware of anyone still using this, thanks for your fix!

PR #918, opened 2023-03-29
gui-apps/xwaylandvideobridge: new package, add 9999

Nowa-Ammerlaan:
Wanted to give this a try, but couldn't quite get it to work. Upstream says: 

> Right now this requires a specific hash of kpipewire until we bump this to Qt6

It does not build with 5.27.3 or 5.27.49.9999 of kpipewire. Presumably it does build with kpipewire-9999, but this I couldn't get to work at the moment because it requires 9999 versions of a bunch of other kde framework stuff. I might give it another shot later.

Possibly this should be in a different category, I put it in `gui-apps` for now since it is not technically limited to kde so the `kde-*` categories wouldn't make much sense.


a17r:
Maybe you can find out what it needs cherry-picking from master branch to work? Upstream will not add new features to 5.27 branch, but we can still decide otherwise depending on the size of it. Otherwise this will take some time before it can be added.

AndrewAmmerlaan:
> Maybe you can find out what it needs cherry-picking from master branch to work? Upstream will not add new features to 5.27 branch, but we can still decide otherwise depending on the size of it. Otherwise this will take some time before it can be added.

Actually, I tried this again with kpipewire-5.27.4-r1 and now it compiles just fine. I updated and simplified the ebuild a bit (there is now a convenient category on invent.kde.org that we can reference), and added a snapshot ebuild of the latest commit at the time of writing.

gui-apps/xwaylandvideobridge/xwaylandvideobridge-9999.ebuild
```diff
@@ -0,0 +1,38 @@
+# Copyright 1999-2023 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+KFMIN=5.83.0
+QTMIN=5.15.0
+KDE_ORG_CATEGORY="system"
+
+inherit ecm kde.org xdg
```
a17r, line 10: you're not using xdg here, right?

gui-apps/xwaylandvideobridge/xwaylandvideobridge-9999.ebuild
```diff
@@ -0,0 +1,38 @@
...
+if [[ ${KDE_BUILD_TYPE} == release ]]; then
+	KEYWORDS="~amd64"
+fi
+
+LICENSE="GPL-2+"
+SLOT="0"
+
+BDEPEND="
+	virtual/pkgconfig
+"
```
a17r, line 23: This can be in one line, should be below `RDEPEND`.

gui-apps/xwaylandvideobridge/xwaylandvideobridge-9999.ebuild
```diff
@@ -0,0 +1,38 @@
+# Copyright 1999-2023 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+KFMIN=5.83.0
+QTMIN=5.15.0
+KDE_ORG_CATEGORY="system"
+
```
a17r, line 9: unnecessary empty line

gui-apps/xwaylandvideobridge/xwaylandvideobridge-9999.ebuild
```diff
@@ -0,0 +1,38 @@
...
+	>=dev-qt/qtquickcontrols-${QTMIN}
+	>=dev-qt/qtx11extras-${QTMIN}
+	>=kde-frameworks/kcoreaddons-${KFMIN}
+	>=kde-frameworks/ki18n-${KFMIN}
+	>=kde-frameworks/kwindowsystem-${KFMIN}
+	>=kde-frameworks/knotifications-${KFMIN}
+	>=kde-frameworks/kwidgetsaddons-${KFMIN}
+	>=kde-plasma/kpipewire-5.27.4
+	media-libs/freetype
+	x11-libs/libxcb
```
a17r, line 36: should it have a slot op?

gui-apps/xwaylandvideobridge/xwaylandvideobridge-9999.ebuild
```diff
@@ -0,0 +1,38 @@
...
+
+DEPEND="
+	>=dev-qt/qtdbus-${QTMIN}
+	>=dev-qt/qtquickcontrols-${QTMIN}
+	>=dev-qt/qtx11extras-${QTMIN}
+	>=kde-frameworks/kcoreaddons-${KFMIN}
+	>=kde-frameworks/ki18n-${KFMIN}
+	>=kde-frameworks/kwindowsystem-${KFMIN}
+	>=kde-frameworks/knotifications-${KFMIN}
+	>=kde-frameworks/kwidgetsaddons-${KFMIN}
```
a17r, line 33: missing `:5` SLOT dep on all of these, more sorting please.

gui-apps/xwaylandvideobridge/xwaylandvideobridge-9999.ebuild
```diff
@@ -0,0 +1,38 @@
...
+	>=kde-plasma/kpipewire-5.27.4
```
a17r, line 34: missing slot dep here too

gui-apps/xwaylandvideobridge/xwaylandvideobridge-9999.ebuild
```diff
@@ -0,0 +1,38 @@
+# Copyright 1999-2023 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+KFMIN=5.83.0
+QTMIN=5.15.0
```
a17r, line 7: kpipewire depends on `QTMIN=5.15.7` and `KFMIN=5.102.0` respectively so it makes no sense to go below those here.

a17r (changes requested):

gui-apps/xwaylandvideobridge/xwaylandvideobridge-9999.ebuild
AndrewAmmerlaan: Yes indeed it should, completely forgot to do the slot operators here.

gui-apps/xwaylandvideobridge/xwaylandvideobridge-9999.ebuild
AndrewAmmerlaan: It installs an icon and desktop file, so I inherit `xdg` to update those databases in postinst and postrm.

a17r: `ecm.eclass` is already doing the magic for you.

AndrewAmmerlaan: > `ecm.eclass` is already doing the magic for you.

Great! I dropped the explicit inherit of `xdg`.

gui-apps/xwaylandvideobridge/xwaylandvideobridge-0_pre20230421.ebuild
```diff
@@ -0,0 +1,38 @@
...
+
+if [[ ${KDE_BUILD_TYPE} == release ]]; then
+	KEYWORDS="~amd64"
+fi
+
+LICENSE="GPL-2+"
+SLOT="0"
+
+DEPEND="
+	>=dev-qt/qtcore-${QTMIN}:5
```
a17r, line 22: This is provided by ecm.eclass already.

a17r (approved):
Approved after last small fix.

a17r:
@AndrewAmmerlaan please make sure not to needlessly add subslot ops on KF5 deps as they are only for private API usage (see description in metadata.xml).

PR #919, opened 2023-04-02
kde-plasma/discover: add flag for enabling snap backend

AndrewAmmerlaan:
This adds a flag to discover to build the snap backend and adds the library we need as intermediary. 

@gentoo/kde Please let me know if this is conceptually OK. If it is, I'll merge the `sys-libs/snapd-glib` library to ::gentoo with appropriate use masks for `kde-plasma/discover snap` (snap is only available on amd64 with systemd, so the flag needs to be globally masked, and unmasked only on the `default/linux/amd64/*/systemd/*` profiles). The changes to `kde-plasma/discover` can be merged here and can trickle down to ::gentoo in the next release.

Still TODO is to maybe at some point in the future add a flag for the PackageKit backend. This is a bit more work since PackageKit's portage backend needs to be fixed-up first (bump to python3 and migrated away from layman). Though to be honest I am not yet fully convinced this is worth the effort.

Also CC'ing the maintainer of `app-containers/snapd` @zmedico

### DIFF
--- a/kde-plasma/discover/discover-5.27.49.9999.ebuild
+++ b/kde-plasma/discover/discover-5.27.49.9999.ebuild
@@ -15,7 +15,7 @@ HOMEPAGE="https://userbase.kde.org/Discover"
 LICENSE="GPL-2" # TODO: CHECK
 SLOT="5"
 KEYWORDS=""
-IUSE="+firmware flatpak telemetry webengine"
+IUSE="+firmware flatpak snap telemetry webengine"
 
 # libmarkdown (app-text/discount) only used in PackageKitBackend
 DEPEND="
@@ -45,11 +45,13 @@ DEPEND="
 	>=kde-frameworks/purpose-${KFMIN}:5
 	firmware? ( >=sys-apps/fwupd-1.5.0 )
 	flatpak? ( sys-apps/flatpak )
+	snap? ( sys-libs/snapd-glib:=[qt5] )
 	telemetry? ( dev-libs/kuserfeedback:5 )
 	webengine? ( >=dev-qt/qtwebview-${QTMIN}:5 )
 "
 RDEPEND="${DEPEND}
 	>=dev-qt/qtquickcontrols2-${QTMIN}:5
+	snap? ( app-containers/snapd )
 "
 BDEPEND=">=kde-frameworks/kcmutils-${KFMIN}:5"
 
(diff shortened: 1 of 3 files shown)


kde-plasma/discover/discover-5.27.49.9999.ebuild
```diff
@@ -67,11 +69,16 @@ src_prepare() {
 
 src_configure() {
 	local mycmakeargs=(
+		# TODO: Port PackageKit's portage back-end to python3 and
+		# move from layman to eselect-repository
 		-DCMAKE_DISABLE_FIND_PACKAGE_packagekitqt5=ON
-		-DCMAKE_DISABLE_FIND_PACKAGE_Snapd=ON
-		-DWITH_KCM=OFF
+		-DWITH_KCM=ON
```
a17r, line 75: Why do we enable that?

https://invent.kde.org/plasma/discover/-/merge_requests/142

a17r (changes requested):
No objections, though reasons need to be given for enabling the KCM.

kde-plasma/discover/discover-5.27.49.9999.ebuild
AndrewAmmerlaan: > No objections, though reasons need to be given for enabling the KCM.

I had enabled this out of curiosity, I thought it might allow for automatically updating snaps and flatpaks. But I have been running with this enabled for a while and this appears to not be the case (I have one flatpak installed and it is not updating automatically). I have disabled this again and added a comment indicating why it is disabled.

a17r (approved):
Thanks, the comment is appreciated so I don't have to look it up again for myself in the future.

kde-plasma/discover/discover-5.27.49.9999.ebuild
```diff
@@ -67,11 +69,18 @@ src_prepare() {
 
 src_configure() {
 	local mycmakeargs=(
+		# TODO: Port PackageKit's portage back-end to python3 and
+		# move from layman to eselect-repository
```
a17r, line 73: Not sure why this comment needs to extend to the details of layman vs. eselect-repository, as unrealistic as it is that any porting of that backend will happen.

AndrewAmmerlaan: Merged with a smaller comment.

a17r:
If you want that flag also in Gentoo ebuild repository then keywording of dependency will be required.

AndrewAmmerlaan:
> If you want that flag also in Gentoo ebuild repository then keywording of dependencies will be required.

AFAIK snap only works on amd64, so keywording on other arches probably doesn't make sense. I added a use flag mask earlier that masks the flag on non-amd64 which should also fix the problem: https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=77564f87710b43364f82f724664d81ea1beb3e10

a17r:
ah, I forgot about that, thanks.

PR #923, opened 2023-05-25
Replace virtualx.eclass with QT_QPA_PLATFORM=offscreen in ecm.eclass

a17r:
Worked with kde-apps/kpat, didn't test more than that, yet.


thesamesam:
Not tested but lgtm. No promises, but Imaybe hope bring this into the Qt stuff for tests like I mentioned the other day (for Qt 6 anyway) at some point (re https://github.com/chimera-linux/cports/pull/175).

a17r:
Thanks, let's just test this live then.

PR #924, opened 2023-08-18
Add kde-frameworks/kwidgetsaddons to xwaylandvideobridge dependencies.

bivsk:
Both ebuilds fail to build without kde-frameworks/kwidgetsaddons. 

Here is the [build log](https://github.com/gentoo/kde/files/12384112/buildlog.txt) when failing.
Output is identical between ebuilds.

Additionally, I've renamed the non 9999 version for a revbump, since there were changes to RDEPEND.


a17r:
Thanks for your contribution. Next time, please start your commit message with `gui-apps/xwaylandvideobridge: `, will fix on merge.

a17r:
OTOH, there's this MR upstream: https://invent.kde.org/system/xwaylandvideobridge/-/merge_requests/14

That would make it obsolete for 9999 and we could just spin a new snapshot.

PR #930, opened 2023-08-27
kde-frameworks/breeze-icons: new '24px' flag

HanabishiRecca:
Makes build-time icons generation optional.

Generation of additional 24x24 icons is the only thing that requires build time processing and additional build deps.
I am not sure about the flag name though.


a17r:
Why not have them - setting aside build time?

a17r:
That's why you can't come up with a fitting use flag name and description so users can decide what they are in for. You want to make a change but have others figure out the implications?

a17r:
It got introduced here: https://invent.kde.org/frameworks/breeze-icons/-/commit/d0582a4f16bdf3947a085c31c5a0136e6283e7a1

PR #931, opened 2023-11-10
gui-apps/xwaylandvideobridge: require kde-frameworks/kwindowsystem[X]

bivsk:
kde-frameworks/kwindowsystem must be built with USE="X", since KX11Extras is not installed without it.


a17r:
RDEPEND changes require revbumps. But I'll fix that for you on merge.

Indeed `src/contentswindow.cpp` makes use of the header. Thanks!

PR #933, opened 2023-11-14
media-video/plasmatube: new package, add 23.08.49.9999, 9999

CyberTailor:


media-video/plasmatube/plasmatube-23.08.49.9999.ebuild
```diff
@@ -0,0 +1,36 @@
...
+	>=kde-frameworks/kconfig-${KFMIN}:5
+	>=kde-frameworks/kcoreaddons-${KFMIN}:5
+	>=kde-frameworks/ki18n-${KFMIN}:5
+	>=kde-frameworks/kirigami-${KFMIN}:5
+	media-video/mpv:=[libmpv]
+"
+RDEPEND="${DEPEND}
+	dev-libs/kirigami-addons:5
+	>=dev-qt/qtgraphicaleffects-${QTMIN}:5
+	>=dev-qt/qtmultimedia-${QTMIN}:5[qml]
```
redstrate, line 34: Technically not needed, we don't use qtmultimedia ourselves but some parts of kirigami-addons might.

redstrate:
23.08 ebuild looks fine, but the 9999 build will fail since master is now Qt6 and the deps changed.

media-video/plasmatube/plasmatube-23.08.49.9999.ebuild
CyberTailor: "src/contents/ui/videoplayer" imports QtMultimedia

CyberTailor:
AFAIK 9999 is not used directly but needed for "regenerate-files" script

a17r:
9999 is used by whomever wants to use 9999, so if you want to add it, it should also have correct dependencies.

media-video/plasmatube/plasmatube-23.08.49.9999.ebuild
redstrate: You're right, that still applies to 23.08 just not in 9999

PR #935, opened 2023-12-07
Gear 24.02 prep

a17r:
TODO:
- net-libs/* multibuild (install image does not work at all right now) @thesamesam maybe you have an idea :)
- media-libs/pulseaudio-qt Qt6-enablement
- kde-apps/kaccounts-integration:6 with "KF5 compat build" providing library components for KF5-based applications, no idea how that will work out
- ~~blockers against SLOT=5 (will add gear.kde.org.eclass mechanism)~~
- any kind of dependency checks (done: kio-extras)
- packaging [mimetreeparser](https://invent.kde.org/pim/mimetreeparser)


net-libs/signon-oauth2/signon-oauth2-9999.ebuild
```diff
@@ -0,0 +1,83 @@
+# Copyright 1999-2023 Gentoo Authors
```
thesamesam, line 1: This one is a mess because `net-libs/signond` installs files referencing Qt 6, like:
```
$ cat /usr/lib64/pkgconfig/signon-plugins.pc
prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}/lib64
includedir=${prefix}/include
plugin_includedir=${includedir}/signon-plugins/SignOn
plugindir=/usr/lib64/signon
Name: signon-plugins
Description: SignonPlugins
Version: 8.61
Requires: libsignon-qt6
Libs.private: -L${libdir} -lQt6Core
Cflags:  -I${includedir}/signon-plugins -I${includedir}/signon-plugins/SignOn
Libs: -lsignon-plugins

$ cat /usr/lib64/pkgconfig/SignOnExtension.pc
prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}/lib64
includedir=${prefix}/include
plugindir=/usr/lib64/signon/extensions

Name: SignOnExtension
Description: SignOn extension development
Version: 8.61
Requires: Qt6Core
Libs: -L${libdir} -lsignon-extension
Cflags: -I${includedir}/signon-extension
```

thesamesam: The problem is, even if say, we make the .pc files prefixed/suffixed with the Qt version, that makes us incompatible with other distros and is very ungentoo.

So we're down to:
1) Do upstream care about co-installability here?
2) Can we get away with only allowing qt5 or qt6?

thesamesam: thanks, done at https://gitlab.com/accounts-sso/signon-plugin-oauth2/-/merge_requests/28#note_1687358193

thesamesam: He's replied so I'll try look at that tomorrow.

thesamesam: https://github.com/thesamesam/kde/tree/gear-24.02-prep may be good enough here now. The only real question in my head is whether we should try to split signond (hopefully not) but I _think_ the dependencies enforcing the matching Qt version in plugins are enough.

thesamesam:
Merged.

PR #940, opened 2023-12-09
Add switch to co-install KF5 KIO with Plasma 6 kio-extras

kingfared:


a17r (changes requested):
You are making changes to 2 different packages, they need to be 2 separate commits.
Please also add the upstream commit reference introducing that option to KF5KIO in git summary.
Also, the new (local) IUSE will need use flag metadata added to metadata.xml, just copy from existing examples in kde overlay.
In addition it would be nice to mention the reason for this blocker in kio-extras git summary, the import of `kcm_trash` from KF5KIO (see https://invent.kde.org/network/kio-extras/-/commit/edae36345810fef100bbb7f09c5bb44f8d0d9fea). Actually all of the KCMs residing kio-extras' kcms subdirectory seem to have been imported from KIO 7 months ago.

Thanks!

kingfared (commented):
Patch split and added metadata for new flag.

a17r (approved):
Thanks!

a17r:
minor nit, fixing on merge: kio-extras is not part of Plasma 6.

PR #941, opened 2023-12-11
kde-apps/filelight: Add missing kirigami-addons dependency

redstrate:
Was recently added in the Qt6 port, is normally a runtime dependency but in this case is also added to the CMakeLists.


a17r (changes requested):
Please keep dependencies sorted, alphanumerically (style guide: unconditional dependency block above conditional, if there are use flags).

Thanks!

redstrate:
> Please keep dependencies sorted, alphanumerically (style guide: unconditional dependency block above conditional, if there are use flags).
> 
> Thanks!

Fixed

PR #942, opened 2023-12-11
kde-plasma/plasma-sdk: Add missing plasma5support dependency

redstrate:
This seems to be a new one, and it's enforced by CMake


thesamesam:
would you mind including a link to the upstream commit in the commit msg too? thank you!

redstrate:
> would you mind including a link to the upstream commit in the commit msg too? thank you!

Sure thing! Added now, not sure if you have a preferred format for that

PR #943, opened 2023-12-11
kde-plasma/xdg-desktop-portal-kde: Add dependency on qtbase with cups

redstrate:
xdg-desktop-portal-kde depends on printsupport (and a private header of it, no less) so make sure we depend on qtbase having the cups USE flag enabled.

### DIFF
--- a/kde-plasma/xdg-desktop-portal-kde/xdg-desktop-portal-kde-9999.ebuild
+++ b/kde-plasma/xdg-desktop-portal-kde/xdg-desktop-portal-kde-9999.ebuild
@@ -16,10 +16,11 @@ SLOT="6"
 KEYWORDS=""
 IUSE=""
 
+# dev-qt/qtbase:=[cups]: includes specifically the cups private header
 # dev-qt/qtgui: QtXkbCommonSupport is provided by either IUSE libinput or X
 COMMON_DEPEND="
 	>=dev-libs/wayland-1.15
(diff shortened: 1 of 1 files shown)


a17r:
Are you sure that is needed? According to ebuild, QtPrintSupport is enabled by USE=gui,widgets - not necessarily implying cups support.

thesamesam:
ah, I remember making that mistake before, ugh

a17r:
If indeed it includes a private header, we should add := slot op to qtbase.

kde-plasma/xdg-desktop-portal-kde/xdg-desktop-portal-kde-9999.ebuild
```diff
@@ -16,10 +16,11 @@ SLOT="6"
 KEYWORDS=""
 IUSE=""
 
+# dev-qt/qtbase:=[cups]: includes specifically the cups private header
 # dev-qt/qtgui: QtXkbCommonSupport is provided by either IUSE libinput or X
 COMMON_DEPEND="
 	>=dev-libs/wayland-1.15
-	>=dev-qt/qtbase-${QTMIN}:6[dbus,gui,widgets]
+	>=dev-qt/qtbase-${QTMIN}:=[cups,dbus,gui,widgets]
```
thesamesam, line 23: we can do :6=

PR #944, opened 2023-12-11
kde-plasma/plasma-nm: Remove patch that was upstreamed

redstrate:
Introduced in commit a0e80449ec9209010354b4af8aa30b7563c98436.


a17r:
Patch was not really upstreamed, btw, it was just conflicting. But following up on that: https://github.com/gentoo/kde/pull/944

PR #946, opened 2023-12-13
oxygen-9999.ebuild add qt5 useflag

glitchking007:
Based on upstream the default is to build both qt5 and qt6. adding the ability to just build qt6.


a17r:
Thanks for your contribution. Please squash commits into one and adjust git summary to Gentoo standards: https://www.gentoo.org/glep/glep-0066.html#commit-messages

PR #945, opened 2023-12-11
dev-libs/mpvqt: new package, add 1.0.0

redstrate:
MpvQt was finally released (yay!) and this will be needed by Tokodon in 9999/24.02, and also PlasmaTube 9999/24.02 (#933)

PR #952, opened 2023-12-16
media-video/plasmatube: new package, add 23.08.49.9999, 9999

redstrate:
Supersedes #945. I tested both the 23.08 and 9999 versions and the dependencies should be correct now.


redstrate:
Oh and #945 should be merged first, of course

PR #953, opened 2023-12-16
net-im/tokodon: Update dependencies and move to mpvqt package

redstrate:
Added some missing dependencies I noticed, and shuffled some which should be a RDEPEND (tell me if that's actually wrong!) Also moved to the new mpvqt package in #945 which is what's required now


a17r (approved):
Thanks a lot!

PR #956, opened 2023-12-18
kde-plasma/discover: appstream version 1.0 is required

rachid-a:


rachid-a:
There is request for updating appstream to version 1.0.0 https://bugs.gentoo.org/915503

a17r:
> There is request for updating appstream to version 1.0.0 https://bugs.gentoo.org/915503

Yes, but it is a packaging hassle and will introduce a "can't update" conflict as soon as pushed to tree.

PR #960, opened 2023-12-18
Add missing remote-id's

parona-source:
Some low hanging fruit.

This is just what pkgcheck reports, so packages without non live packages are missed (pkgcheck only checks SRC_URI and HOMEPAGE for the check).

### DIFF
--- a/app-admin/systemdgenie/metadata.xml
+++ b/app-admin/systemdgenie/metadata.xml
@@ -13,5 +13,6 @@
 			<email>rthomsen6@gmail.com</email>
 			<name>Ragnar Thomsen</name>
 		</maintainer>
+		<remote-id type="kde-invent">system/systemdgenie</remote-id>
 	</upstream>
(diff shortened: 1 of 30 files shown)


a17r:
lgtm, thanks for doing this - still working on it (hence draft)?

parona-source:
Just removed the bugs-to as ksanescore didnt have product page there.

Also I guess NonexistentBlocker would be a good candidate to ignore globally.

a17r:
Do you have some automatism for this? If yes, you could locally temporarily change *.kde.org eclasses to always set HOMEPAGE and gobble up results.

parona-source:
I could, ideally pkgcheck would be able to use EGIT_REPO_URI but that may be more involved.

a17r:
> Also I guess NonexistentBlocker would be a good candidate to ignore globally in pkgcheck.

That's actually a pkgcheck bug when you're in a branch.

PR #961, opened 2023-12-18
kde-plasma/qqc2-breeze-style: new package, add 9999

parona-source:
Noticed talk on irc, so here is the package. @redstrate

### DIFF
--- a/kde-plasma/qqc2-breeze-style/metadata.xml
+++ b/kde-plasma/qqc2-breeze-style/metadata.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "https://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="project">
+		<email>kde@gentoo.org</email>
+		<name>Gentoo KDE Project</name>
+	</maintainer>
+	<upstream>
+		<bugs-to>https://bugs.kde.org/</bugs-to>
+		<remote-id type="kde-invent">plasma/qqc2-breeze-style</remote-id>
+	</upstream>
+</pkgmetadata>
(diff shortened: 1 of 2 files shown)


kde-plasma/qqc2-breeze-style/metadata.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "https://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="project">
+		<email>kde@gentoo.org</email>
+		<name>Gentoo KDE Project</name>
+	</maintainer>
+	<upstream>
+		<bugs-to>https://bugs.kde.org/</bugs-to>
```
redstrate, line 9: I think we are using the Invent bug tracker, or at least not Bugzilla. Or at least it didn't have a product the last time I checked...

a17r:
Do we depend on this from anywhere?

redstrate:
> Do we depend on this from anywhere?

Not yet, but plasma-integration has begun to start enforcing this for QGuiApplications if they don't set one specifically so some applications (a very small portion of them, to be frank) will start using this style.

a17r (commented):
@redstrate: this seems like an upstream leftover since it does not exist in Qt6:
```
-- The following OPTIONAL packages have not been found:

 * Qt6X11Extras (required version >= 6.6.0)
```

kde-plasma/qqc2-breeze-style/metadata.xml
a17r: That would normally be reserved for dev bugs, not user bugs. Maybe something that needs to be requested from KDE infra.

parona-source:
https://doc.qt.io/qt-6/extras-changes-qt6.html#changes-to-qt-x11-extras

"The [Qt X11 Extras](https://doc.qt.io/qt-5/qtx11extras-index.html) module was removed in Qt 6."

Yep.

a17r (approved):
Thanks!

PR #964, opened 2023-12-22
kde-misc/kio-fuse: add 5.1.0

thesamesam:


thesamesam:
Merged (to ::gentoo, but with ::kde live sync).

PR #965, opened 2023-12-22
Workaround Portage getting too confused by blockers

a17r:


thesamesam (approved):
Thanks.

Please add:
```
Bug: https://bugs.gentoo.org/813099
Bug: https://bugs.gentoo.org/814335
```

PR #966, opened 2023-12-23
kde-plasma/kwin: require wayland to be enabled in KF dependencies

parona-source:
* Kwin uses wayland unconditionally and expects wayland headers which are available only with the wayland use flags.


parona-source:
https://github.com/gentoo/kde/pull/967

a17r:
This is again a case of pkgcheck bug because you're in a branch. This package has been in kde overlay.

parona-source:
https://github.com/gentoo/kde/pull/963 This should address the cases where it was in the kde overlay.

thesamesam:
rebase?

PR #967, opened 2023-12-23
.github/workflows: ignore more keywords for merge requests

parona-source:
* Kde overlay adds blockers for packages that are outside the repo, which means false positives.


a17r:
otoh we may want to know when we can drop blockers, so we may just need to live with some false warnings?

parona-source:
Would you be fine it its only ignored for merge request?

a17r:
If that's possible?

parona-source:
Modified it to do that. Specifying keywords on the command leads to to the configured keywords to be ignored, so all of them have to be specified.

PR #969, opened 2023-12-27
media-libs/pulseaudio-qt6: Add Qt6 support

aidanharris:


media-libs/pulseaudio-qt6/pulseaudio-qt6-9999.ebuild
```diff
@@ -0,0 +1,37 @@
...
+if [[ ${KDE_BUILD_TYPE} = release ]]; then
+	SRC_URI="mirror://kde/stable/${MY_PN}/${MY_P}.tar.xz"
+	KEYWORDS="~amd64 ~arm64 ~ppc64 ~riscv ~x86"
+fi
+
+LICENSE="LGPL-2.1"
+SLOT="0/3"
+
+RDEPEND="
+	>=dev-qt/qtgui-${QTMIN}:6
```
thesamesam, line 29: See https://invent.kde.org/libraries/pulseaudio-qt/-/blob/master/CMakeLists.txt?ref_type=heads#L40 vs https://wiki.gentoo.org/wiki/Project:Qt/Qt6_migration_notes.

aidanharris: Thanks. I didn't know that page existed.

thesamesam: It's new :D

a17r:
kde-misc/kdeconnect still is the only revdep of media-libs/pulseaudio-qt. That means a new package name just for it being ported to Qt6 is not justifiable. Let's pin down kdeconnect:5 to <media-libs/pulseaudio-qt-1.4 instead. But we do not even have a release of that yet, as kdeconnect is not yet officially shipped as Qt6-based in 24.02:

https://community.kde.org/KDE_Gear/24.02_Release_notes#Tarballs_have_switched_from_being_Qt5_based_to_Qt6-based

That means we are not going to switch kdeconnect to KF6 yet, either. I'm going to ask upstream about it tomorrow though.

media-libs/pulseaudio-qt6/pulseaudio-qt6-9999.ebuild
a17r: Regardless, you would have noticed upon pkgcheck, which is a bare minimum before filing a PR.

jospezial:
Next 24.xx tag will have it.
https://invent.kde.org/network/kdeconnect-kde/-/commit/7f03aa548c233ab44d1cbb12f3ced8303ce95b9c
"Fully port to Qt6"

https://invent.kde.org/network/kdeconnect-kde/-/merge_requests/611
was merged also.

a17r:
very good.

PR #971, opened 2023-12-28
kde-plasma/discover, media-gfx/digikam: remove unused patch, fix patchname

mm1ke:
Hi,

Two little fixes regarding patches.


a17r (approved):
Thanks!

PR #973, opened 2024-01-03
cmake.eclass settings

a17r:


eclass/cmake.eclass
```diff
@@ -458,6 +458,8 @@ cmake_src_configure() {
 		set(CMAKE_INSTALL_DOCDIR "${EPREFIX}/usr/share/doc/${PF}" CACHE PATH "")
 		set(BUILD_SHARED_LIBS ON CACHE BOOL "")
 		set(Python3_FIND_UNVERSIONED_NAMES FIRST CACHE STRING "")
+		set(FETCHCONTENT_FULLY_DISCONNECTED ON CACHE BOOL "")
```
thesamesam, line 461: Maybe we should check if PROPERTIES=live and only add this if it's unset? (to not mess with live ebuilds)

a17r: Why would we make a difference for live sources?

thesamesam: I suppose because people may expect live ebuilds to be unsandboxed network-wise and hence be able to fetch anything. But I'm fine with trying it as-is and seeing.

thesamesam (approved):

eclass/cmake.eclass
ionenwks: The only ebuilds this would affect is overlays asking users to actually turn off network-sandbox, not like cmake could fetch anything even in a live ebuild unless it's called in src_unpack.

At that point I'd say they should define the option manually in mycmakeargs.

ionenwks (approved):
And thanks for this, was tired of these in ebuilds myself.

PR #975, opened 2024-01-07
Documentation: lib.sh: Support slot scope in get_main_tree_keyword()

a17r:
In general, use more portageq instead of sed, grep, ls, sort and tail.


ionenwks (approved):
Not familiar with these scripts but I don't see anything that stands out as bad in the changes from a quick look.

thesamesam (approved):

a17r:
Needs more work for the case of no ebuild in main tree available.

PR #979, opened 2024-02-12
Add Microsoft GSL to allow kio-extras-9999 to compile

FireBurn:
kio-extras-9999 has a new dependency


a17r:
It would be nice to always include the upstream commit reference to when this was added. So I had to look it up myself, and found that use of this is limited to IUSE=sftp. The dependency was met with some opposition in the MR: https://invent.kde.org/network/kio-extras/-/merge_requests/326

Chiitoo:
There seems to be a 'dev-cpp/ms-gsl' already (renamed from 'dev-cpp/gsl' back in 2018), which looks to be the same thing.

FireBurn:
Fantasic, I'll get this updated and to only use the it if USE=sftp is set

a17r:
ping :)

PR #982, opened 2024-03-17
kde-apps/gwenview: correct package KDcraw package name

parona-source:


kde-apps/gwenview/gwenview-9999.ebuild
```diff
@@ -45,8 +45,8 @@ COMMON_DEPEND="
 	>=kde-frameworks/kxmlgui-${KFMIN}:6
 	>=kde-frameworks/solid-${KFMIN}:6
 	media-gfx/exiv2:=
-	>=media-libs/kcolorpicker-0.3.0
-	>=media-libs/kimageannotator-0.7.0
+	>=media-libs/kcolorpicker-0.3.0:=
+	>=media-libs/kimageannotator-0.7.0:=
```
a17r, line 49: There is no subslot to trigger rebuild against these packages, last time I checked, so this will do nothing.

Since gwenview:5 depends firmly on <0.3.0 and <0.7.0 here, and only (still masked) gwenview:6 on these newer versions, simply raise the minimum required versions to 0.3.1 and 0.7.1 respectively to fix upstream's onetime soversion glitch.

a17r (changes requested):

kde-apps/gwenview/gwenview-9999.ebuild
a17r: You're leaving the minimum version req bump for me to do then? ;)

PR #984, opened 2024-03-23
kde-plasma/plasma-desktop: install our own mimeapps.list

thesamesam:
Start by importing kde-plasma/plasma-desktop's own mimeapps.list and updating
the list in subsequent commits for easier review.

Bug: https://bugs.gentoo.org/926412
Signed-off-by: Sam James <sam@gentoo.org>

--
This is still very rough and I need to proof-read it, even. Let alone test it.


a17r:
Would it be beneficial to do this in a standalone package for plasma-desktop to RDEPEND on?

thesamesam:
Ah, that might be a good idea - less hassle when we make changes then, especially as we may make quite a few to start with.

PR #985, opened 2024-04-04
kde-plasma/plasma-pa-9999: Add dependency on media-libs/pulseaudio-qt

FireBurn:
https://invent.kde.org/plasma/plasma-pa/-/merge_requests/248

### DIFF
--- a/kde-plasma/plasma-pa/plasma-pa-9999.ebuild
+++ b/kde-plasma/plasma-pa/plasma-pa-9999.ebuild
@@ -33,6 +33,7 @@ DEPEND="
 	>=kde-plasma/libplasma-${PVCUT}:6
 	media-libs/libcanberra
 	media-libs/libpulse
+	>=media-libs/pulseaudio-qt-${KFMIN}:6
 "
 RDEPEND="${DEPEND}
 	dev-libs/kirigami-addons:6


kde-plasma/plasma-pa/plasma-pa-9999.ebuild
```diff
@@ -33,6 +33,7 @@ DEPEND="
 	>=kde-plasma/libplasma-${PVCUT}:6
 	media-libs/libcanberra
 	media-libs/libpulse
+	media-libs/pulseaudio-qt
```
a17r, line 36: We don't need a slot op?

a17r (changes requested):
Please also change git summary:

`kde-plasma/plasma-pa: Add dependency on media-libs/pulseaudio-qt`

kde-plasma/plasma-pa/plasma-pa-9999.ebuild
FireBurn: I'll add that, but it does require newer than 1.4.0, should I put > 1.4.99 or something like that?

jospezial:
[ebuild  N    *] media-libs/pulseaudio-qt-9999:0/3::kde  USE="-debug -doc -handbook -test" 0 KiB

so :6 is not there.
If 0/3 is the so-version like in pulseaudio-qt-1.3-r4.ebuild then pulseaudio-qt-9999 slot should be bumped to 0/5
and pulseaudio-qt-1.4.0-* 0/4
https://invent.kde.org/libraries/pulseaudio-qt/-/blame/master/CMakeLists.txt?ref_type=heads#L39

kde-plasma/plasma-pa/plasma-pa-9999.ebuild
a17r: `PROJECT_VERSION` in upstream CMakeLists.txt is already bumped to `1.5.0` so use that.

Depending on `${KFMIN}:6` does not work at all right now as @jospezial already pointed out.

a17r: https://invent.kde.org/plasma/plasma-pa/-/merge_requests/254

PR #986, opened 2024-04-06
net-libs/accounts-qt: move away from stale fork

parona-source:
* Branch was removed either way as changes were merged.


a17r:
Thanks, let's move back to stale upstream then. Note that there are some more changes to be done there.

parona-source:
> Thanks, let's move back to stale upstream then. Note that there are some more changes to be done there.

I was under the impression the changes were pushed upstream. The only reason I noticed this was because the qt6 branch was removed.

PR #987, opened 2024-04-07
kde-apps/spectacle: Add OpenCV as a dependency

FireBurn:
https://github.com/KDE/spectacle/commit/0582f205abea33822e765b6afd0b6705078ecad5


a17r:
In general please don't use the github mirror to cite upstream changes. invent.kde.org is official upstream repository.

kde-apps/spectacle/spectacle-9999.ebuild
```diff
@@ -25,6 +25,7 @@ COMMON_DEPEND="
 	>=dev-qt/qtdeclarative-${QTMIN}:6
 	>=dev-qt/qtmultimedia-${QTMIN}:6[qml]
 	>=dev-qt/qtwayland-${QTMIN}:6
+	media-libs/opencv[qt6]
```
a17r, line 28: We don't need a slot operator there?

a17r (changes requested):

PR #989, opened 2024-04-10
kde-plasma/plasma-desktop: Update patch

FireBurn:
https://invent.kde.org/plasma/plasma-desktop/-/merge_requests/2186


FireBurn:
They've changed things again, and this will require more work

PR #991, opened 2024-04-26
kde-plasma/discover: update use dep on snapd-glib

AndrewAmmerlaan:
Version 1.65 of snapd-glib now allows building both for qt5 and qt6, though not at the same time. As per [Qt packaging policy](https://wiki.gentoo.org/wiki/Project:Qt/Policies) I have implemented both a `qt5` and `qt6` flag in this version and we prefer qt5 over qt6 if both flags are enabled.

This implies some changes in the discover:6 ebuild. To get the snapd-glib library we need, we must not only ensure the `qt6` flag is enabled, but also ensure the `qt5` flag is disabled.

### DIFF
--- a/kde-plasma/discover/discover-6.0.49.9999.ebuild
+++ b/kde-plasma/discover/discover-6.0.49.9999.ebuild
@@ -42,7 +42,7 @@ DEPEND="
 	>=kde-frameworks/purpose-${KFMIN}:6
 	firmware? ( >=sys-apps/fwupd-1.9.4 )
 	flatpak? ( sys-apps/flatpak )
-	snap? ( sys-libs/snapd-glib:=[qt6(-)] )
+	snap? ( sys-libs/snapd-glib:=[qt6(-),-qt5(-)] )
 	telemetry? ( >=kde-frameworks/kuserfeedback-${KFMIN}:6 )
 	webengine? ( >=dev-qt/qtwebview-${QTMIN}:6 )
 "
--- a/kde-plasma/discover/discover-9999.ebuild
+++ b/kde-plasma/discover/discover-9999.ebuild
@@ -42,7 +42,7 @@ DEPEND="
 	>=kde-frameworks/purpose-${KFMIN}:6
 	firmware? ( >=sys-apps/fwupd-1.9.4 )
 	flatpak? ( sys-apps/flatpak )
-	snap? ( sys-libs/snapd-glib:=[qt6(-)] )
+	snap? ( sys-libs/snapd-glib:=[qt6(-),-qt5(-)] )
 	telemetry? ( >=kde-frameworks/kuserfeedback-${KFMIN}:6 )
 	webengine? ( >=dev-qt/qtwebview-${QTMIN}:6 )
 "


a17r:
> This implies some changes in the discover:6 ebuild. To get the snapd-glib library we need, we must not only ensure the `qt6` flag is enabled, **but also ensure the `qt5` flag is disabled**.

That's rather weird. Shouldn't snapd-glib ebuild's `REQUIRED_USE` take care of that?

AndrewAmmerlaan:
> > This implies some changes in the discover:6 ebuild. To get the snapd-glib library we need, we must not only ensure the `qt6` flag is enabled, **but also ensure the `qt5` flag is disabled**.
> 
> That's rather weird. Shouldn't snapd-glib ebuild's `REQUIRED_USE` take care of that?

Well the policy explicitly says not to do that, and to simply prefer qt5 over qt6 if both are set.

a17r:
In this case it is not about gui support though, maybe @ionenwks can help clear things up in the spirit of Qt policy.

ionenwks:
> Version 1.65 of snapd-glib now allows building both for qt5 and qt6, though not at the same time.

Have you looked if it's possible to *install* both at same time? I get it doesn't allow building, but that could potentially be solved with multibuild.

Both REQUIRED_USE or [qt6,-qt5] would be rather annoying for users, and I'd rather alternative options be looked at.

ionenwks:
Guess yet another option could be to simply drop support for qt5. Version/revision could potentially stay masked until Plasma 6 is.

ionenwks:
> > Version 1.65 of snapd-glib now allows building both for qt5 and qt6, though not at the same time.
> 
> Have you looked if it's possible to _install_ both at same time? I get it doesn't allow building, but that could potentially be solved with multibuild.
Well, a quick look I do see a single /usr/lib64/libsnapd-qt-2.so.1.0.0 so probably not (was hoping it'd be qt6.so and qt5.so or something).

Guess one more option could be separate revisions for qt5 and qt6 (that block each others), and have discover pull the right one. Not that I looked too closely at this.

ionenwks:
That aside, I'm more against a misleading [qt6,-qt5] than the REQUIRED_USE, but please consider alternatives before jumping on that solution. It should be a last resort.

AndrewAmmerlaan:
> > Version 1.65 of snapd-glib now allows building both for qt5 and qt6, though not at the same time.
> 
> Have you looked if it's possible to _install_ both at same time? I get it doesn't allow building, but that could potentially be solved with multibuild.

This is probably possible (with some patching), but note that Discover is the only reverse dependency of snapd-glib. Nothing else depends on snapd-glib so there is no point in installing both the qt5 and qt6 version at same time, therefore I think that multibuild adds a lot of complexity for no real purpose. I'd rather just drop `USE=qt5` on the newer version, but this will leave discover:5 users stuck on older versions of snapd-glib.


> Guess one more option could be separate revisions for qt5 and qt6 (that block each others), and have discover pull the right one. Not that I looked too closely at this.

> That aside, I'm more against a misleading [qt6,-qt5] than the REQUIRED_USE, but please consider alternatives before jumping on that solution. It should be a last resort.

I'm fine with doing it whatever way KDE/Qt team thinks best, but please then adjust the policy document. The same `[qt6,-qt5]` USE dependency exists in Freecad/VTK/pcl when building with Qt6, if we do it differently here then I think we should also adjust those ebuilds. And I would not have done it this way in the first place if the policy document hadn't explicitly said to not use `REQUIRED_USE`.

Revision slotting is definitely possible, in fact that is the way I was doing it before upstream merged Qt6 support as a separate option to Qt5 support.

ionenwks:
[qt6,-qt5] may as well be the same as REQUIRED_USE, this is just "working around" the policy

ionenwks:
Ultimately I do tend to feel that packages that cannot support both at same time should just hard-switch to qt6 and move on. Not so great for integration until Plasma 6 is unmasked but well.

AndrewAmmerlaan:
> [qt6,-qt5] may as well be the same as REQUIRED_USE, this is just "working around" the policy

I agree, but it is the logical result when a library with optional support for one of two Qt versions prefers Qt5 over Qt6 when both flags are set and cannot use REQUIRED_USE (which is the current wording of the policy).

I'm not the only one that is doing it this way because the policy says so, see `VTK[qt6]`: https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=d46cf053da6ab5b3ff07979c6fee2e15dc0329e4
Because VTK ignores `USE=qt6` if we also have `USE=qt5`, I was left with no choice but to depend on `vtk[qt6,-qt5]` in `Freecad[qt6]`.

As I said, I'm fine with doing it differently, but then please revise the policy to say something about "libraries that have reverse dependencies and optional support for one of two Qt versions". Otherwise people will continue to do it this way.

ionenwks:
> This is probably possible (with some patching), but note that Discover is the only reverse dependency of snapd-glib. Nothing else depends on snapd-glib so there is no point in installing both the qt5 and qt6 version at same time, therefore I think that multibuild adds a lot of complexity for no real purpose. I'd rather just drop `USE=qt5` on the newer version, but this will leave discover:5 users stuck on older versions of snapd-glib.

Not necessarily, this was the idea with different revisions (one for qt5, and one for qt6). Albeit it helps to have something to identify when doing that (like snapd-glib:0/6 subslot, or fake slots).

ionenwks:
> As I said, I'm fine with doing it differently, but then please revise the policy to say something about "libraries that have reverse dependencies and optional support for one of two Qt versions". Otherwise people will continue to do it this way.

fwiw the current examples do not consider a *library* that can only support one, and the *do not* are cited for end-user application examples.

The top rule is:
`Avoiding unnecessary REQUIRED_USE is important to minimise user interaction, especially as USE="qt5 qt6" becomes a common configuration.`

Unnecessary implies that there can be case where it's necessary, and it's not banned per-se. Just that you should look at other options first.

ionenwks:
> > As I said, I'm fine with doing it differently, but then please revise the policy to say something about "libraries that have reverse dependencies and optional support for one of two Qt versions". Otherwise people will continue to do it this way.
> 
> fwiw the current examples do not consider a _library_ that can only support one, and the _do not_ are cited for end-user application examples.

Oh wait, I missed the last one (I did not write these). Then again, it does not say *do not* for that one still.

ionenwks:
Either way, I'll still add a extra note to that last example to clarify.

ionenwks:
> Either way, I'll still add a extra note to that last example to clarify.

Hope this will do, see 2nd part of the example which formerly only mentioned non-libraries:
https://wiki.gentoo.org/wiki/Project:Qt/Policies#Requires_one_of_two_Qt_versions

AndrewAmmerlaan:
> > Either way, I'll still add a extra note to that last example to clarify.
> 
> Hope this will do, see 2nd part of the example which formerly only mentioned non-libraries: https://wiki.gentoo.org/wiki/Project:Qt/Policies#Requires_one_of_two_Qt_versions

Great! This should help I think. 

@negril Based on the above, could you please adjust `sci-libs/vtk` (and maybe `sci-libs/pcl` too) so we can avoid the weird `[qt6,-qt5]` in `media-gfx/freecad`? Given that freecad requires pyside, and pyside2 is unmaintained upstream, maybe it makes sense to simply switch freecad/vtk/pcl to qt6-only? Though I have not yet checked if there are other reverse dependencies that might require the qt5 versions of vtk and/or pcl (plus qt6 is not yet keyworded everywhere).

I'll close this, and adjust `sys-libs/snapd-glib` later.

PR #996, opened 2024-06-23
kde-plasma/plasma-meta: allow xdg-desktop-portal-gnome with USE=-unsu…

thesamesam:
…pported

USE=-unsupported (the default) banning coinstalls with sys-apps/xdg-desktop-portal-gnome
caused issues with users either migrating between KDE Plasma and GNOME or
wanting to have both installed, so it prompted me to look into it again.

xdg-desktop-portal-kde installs kde-portals.conf which is used and preferred
within the Plasma desktop, so we have no issue here. (Done in upstream
commit ae9b326e55ca4a25a762cf0bfe96f5e947765833).

The recommendation to not coinstall them was added upstream in April 2023
at https://community.kde.org/index.php?title=Distributions/Packaging_Recommendations&diff=prev&oldid=96239,
while the aforementioned upstream commit is from May 2023.

Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/kde-plasma/plasma-meta/plasma-meta-6.1.49.9999.ebuild
+++ b/kde-plasma/plasma-meta/plasma-meta-6.1.49.9999.ebuild
@@ -107,7 +107,6 @@ RDEPEND="
 	thunderbolt? ( >=kde-plasma/plasma-thunderbolt-${PV}:${SLOT} )
 	!unsupported? (
 		!gui-apps/qt6ct
-		!sys-apps/xdg-desktop-portal-gnome
 	)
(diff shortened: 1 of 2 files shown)


thesamesam:
I've asked Nate on IRC whether my understanding is correct.

thesamesam:
Someone pointed out on IRC that it should have worked before the config file, even, because it would inspect `${XDG_CURRENT_DESKTOP}`, but they can't remember / don't know what issues or conflicts existed.

Let's see what others say but I'm inclined to say we go ahead unless we can find some specifics showing problems.

PR #997, opened 2024-06-30
Cleanup KAccounts situation + phonon backend IUSE change

a17r:
Summary:

kde-apps/kaccounts-providers: Drop IUSE +webengine, add qml USEdep
    
    Let's not kid ourselves: net-libs/signon-ui will pull in dev-qt/qtwebengine
    anyway.
    
kde-apps/kaccounts-integration: Add PDEPEND=kde-apps/kaccounts-providers
    
    kde-apps/kaccounts-integration provides kcm_accounts for systemsettings,
    however without kde-apps/kaccounts-providers the list is practically empty
    which makes no sense.
    
    So far, only kde-apps/kdenetwork-meta and kde-misc/kio-gdrive had been
    pulling it in but we cannot assume every Plasma user would have those
    installed.

kde-apps/kaccounts-integration: RDEPEND on signon-kwallet-extension:6
    
    This package already depends on kde-frameworks/kwallet so in this case
    we do not introduce a perceived extra dependency. It is also where Arch
    Linux chose to put it.
    
    Closes: https://bugs.gentoo.org/642420

Various packages: Rename IUSE kaccounts -> webengine
    
    kde-apps/kaccounts-integration inevitably pulls in -providers which
    pulls in dev-qt/qtwebengine.

media-libs/phonon: Switch IUSE +vlc to minimal
    
    With only a single available backend left, it does not make that much sense
    to rely on a commonly disabled flag to provide major functionality, even if
    enabled by default.
    
    Bug: https://bugs.gentoo.org/935033

### DIFF
--- a/kde-apps/akonadi/akonadi-24.05.49.9999.ebuild
+++ b/kde-apps/akonadi/akonadi-24.05.49.9999.ebuild
@@ -17,7 +17,7 @@ HOMEPAGE="https://community.kde.org/KDE_PIM/akonadi"
 LICENSE="LGPL-2.1+"
 SLOT="6"
 KEYWORDS=""
-IUSE="+kaccounts +mysql postgres sqlite tools xml"
+IUSE="+mysql postgres sqlite tools +webengine xml"
 
 REQUIRED_USE="|| ( mysql postgres sqlite ) test? ( tools )"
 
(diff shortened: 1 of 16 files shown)


a17r:
TODO: Integrate "optional" at runtime `kde-apps/signon-kwallet-extension` somehow (re: https://bugs.gentoo.org/642420), maybe in plasma-meta where it does not hurt?

PR #999, opened 2024-08-11
New: ecm-common.eclass and split off several ${PN}-common packages where needed

a17r:
### ecm-common.eclass - new eclass review! ###

What does it do?
- [x] Work with KF5 or KF6 in the background, so BDEPENDs are defined as two blocks inside `|| ( )`
- [x] Create a very simple root `CMakeLists.txt` in place of upstream's
- [x] KDE projects are *very* standardised, so a few lines of CMake will already provide us with translations, docs, translated docs for 95% of packages.
- [x] On top of that, ebuild consumers can define icons and files to install
- [x] ... or inject their own Heredoc when ecm-common.eclass is not enough

What's the goal?
- Easily create ${PN}-common packages for other split ${PN} to depend upon
- Be able to depend on ${PN}-common:0 packages from e.g. both SLOT 5 and 6
- ... cause less friction with not having to set compatibility USE flags and dropping many blockers

When thought through to the end, it would enable us move *all* docs (currently USE=handbook in KDE packages) and translations into separate packages, in the future.


kde-apps/libkcddb/libkcddb-9999.ebuild
```diff
@@ -31,19 +31,26 @@ DEPEND="
...
 	local mycmakeargs=(
 		$(cmake_use_find_package musicbrainz MusicBrainz5)
 	)
-
 	ecm_src_configure
 }
+
+src_install() {
+	ecm_src_install
+	# Shipped by kde-apps/libkcddb-common package for shared use w/ SLOT 5
```
thesamesam, line 53: ${ED}

kde-apps/libksane/libksane-9999.ebuild
```diff
@@ -24,12 +24,23 @@ DEPEND="
...
 src_configure() {
 	local mycmakeargs=(
 		$(cmake_use_find_package kwallet KF6Wallet)
 	)
 	ecm_src_configure
 }
+
+src_install() {
+	ecm_src_install
+	# Shipped by kde-apps/libksane-common package for shared use w/ SLOT 5
```
thesamesam, line 44: ${ED}, die

kde-frameworks/purpose-common/purpose-common-9999.ebuild
```diff
@@ -0,0 +1,41 @@
...
+
+ECM_INSTALL_ICONS=(
+	src/plugins/reviewboard/icons/16-apps-reviewboard-purpose6.png:\${KDE_INSTALL_ICONDIR}
+	src/plugins/reviewboard/icons/128-apps-reviewboard-purpose6.png:\${KDE_INSTALL_ICONDIR}
+	src/plugins/phabricator/icons/16-apps-phabricator-purpose6.png:\${KDE_INSTALL_ICONDIR}
+	src/plugins/phabricator/icons/128-apps-phabricator-purpose6.png:\${KDE_INSTALL_ICONDIR}
+)
+
+ecm-common_inject_heredoc() {
+	if use webengine; then
```
thesamesam, line 30: If you're using `<<-`, it'll strip leading whitespace, so you can indent the block below to look prettier.

kde-frameworks/purpose/purpose-5.239.9999.ebuild
```diff
@@ -66,15 +69,13 @@ src_configure() {
...
-	if use kf6compat; then
-		rm "${D}"/usr/share/icons/hicolor/128x128/apps/{reviewboard,phabricator}-purpose.png \
-			"${D}"/usr/share/icons/hicolor/16x16/apps/{reviewboard,phabricator}-purpose.png \
-			|| die
-		if use kaccounts; then
-			rm "${D}"/usr/share/accounts/services/kde/{google-youtube,nextcloud-upload}.service \
-			|| die
-		fi
+	# Shipped by kde-frameworks/purpose-common package for shared use w/ SLOT 5
+	rm "${D}"/usr/share/icons/hicolor/128x128/apps/{reviewboard,phabricator}-purpose.png \
```
thesamesam, line 73: ED ED ED

kde-apps/kate-common/kate-common-9999.ebuild
```diff
@@ -0,0 +1,33 @@
...
+	!<kde-apps/kate-24.07.90-r1
+	!<kde-apps/kate-addons-24.07.90-r1
+	!<kde-apps/kate-lib-24.07.90-r1
+"
+
+src_prepare() {
+	ecm-common_src_prepare
+
+	# delete colliding kwrite translations
+	find po -type f -name "*po" -and -name "kwrite*" -delete || die
```
thesamesam, line 24: Why doesn't ecm_punt_po_install work here?

eclass/ecm-common.eclass
```diff
@@ -0,0 +1,188 @@
...
+	[[ ${ECM_INSTALL_FILES@a} == *a* ]] ||
+		die "ECM_INSTALL_FILES must be an array"
+fi
+
+# @FUNCTION: ecm-common_inject_heredoc
+# @DESCRIPTION:
+# Override this to inject custom Heredoc into the root CMakeLists.txt
+ecm-common_inject_heredoc() {
+	debug-print-function ${FUNCNAME} "$@"
+	# put your stuff here
```
thesamesam, line 104: ?

kde-apps/kate-common/kate-common-9999.ebuild
a17r: In this case we pretend kwrite is still like an independent package. Otoh, it itself depends on kate-lib anyway, so we might as well do the new standard treatment...

eclass/ecm-common.eclass
```diff
@@ -0,0 +1,236 @@
...
+	for i in "${ECM_INSTALL_FILES[@]}"; do
+		cat >> CMakeLists.txt <<- _EOF_ || die
+		install(FILES ${i%:*} DESTINATION ${i#*:})
+		_EOF_
+	done
+}
+
+# @FUNCTION: ecm-common_inject_heredoc
+# @DESCRIPTION:
+# Override this to inject custom Heredoc into the root CMakeLists.txt
```
thesamesam, line 194: Cute :)

eclass/ecm-common.eclass
```diff
@@ -0,0 +1,236 @@
...
+
+	feature_summary(WHAT ALL INCLUDE_QUIET_PACKAGES FATAL_ON_MISSING_REQUIRED_PACKAGES)
+	_EOF_
+}
+
+# @FUNCTION: ecm-common_src_prepare
+# @DESCRIPTION:
+# Wrapper for cmake_src_prepare with a Heredoc replacing the standard
+# root CMakeLists.txt file to only generate and install translations.
+ecm-common_src_prepare() {
```
thesamesam, line 214: I find the whitespace between lines ugly here but you're the one doing most of the work, so wfm

kde-apps/libkcddb-common/libkcddb-common-9999.ebuild
```diff
@@ -0,0 +1,52 @@
...
+"
+BDEPEND="kde-frameworks/kcmutils:*"
+
+ECM_INSTALL_FILES=(
+	libkcddb/libkcddb5.kcfg:\${KDE_INSTALL_KCFGDIR}
+)
+
+ecm-common_inject_heredoc() {
+	if use handbook; then
+cat <<-EOF >> CMakeLists.txt
```
thesamesam, line 26: No indents here?

kde-apps/libkcddb-common/libkcddb-common-9999.ebuild
```diff
@@ -0,0 +1,52 @@
...
+}
+
+src_prepare() {
+	# prepare to use kcmutils_generate_desktop_file
+	mv kcmcddb/kcm_cddb.json ./ || die
+	ecm-common_src_prepare
+}
+
+src_configure() {
+	if has_version "kde-frameworks/kcmutils:6"; then
```
thesamesam, line 48:
```suggestion
	if has_version -b "kde-frameworks/kcmutils:6"; then
```

kde-frameworks/purpose/purpose-9999.ebuild
```diff
@@ -57,6 +57,18 @@ src_configure() {
 	ecm_src_configure
 }
 
+src_install() {
+	ecm_src_install
+	# Shipped by kde-frameworks/purpose-common package for shared use w/ SLOT 5
+	rm "${ED}"/usr/share/icons/hicolor/128x128/apps/{reviewboard,phabricator}-purpose.png \
```
thesamesam, line 63: I don't love this but it's not that bad.

thesamesam (approved):

eclass/ecm-common.eclass
a17r: me too.

eclass/ecm-common.eclass
```diff
@@ -0,0 +1,328 @@
...
+
+BDEPEND+=" || ( ( ${_KF6_BDEPEND[*]} ) ( ${_KF5_BDEPEND[*]} ) )"
+
+# @FUNCTION: _ecm-common_preamble
+# @INTERNAL
+# @DESCRIPTION:
+# Create a CMakeLists.txt file with minimum ECM setup.
+_ecm-common_preamble() {
+	cat > CMakeLists.txt <<- _EOF_ || die
+	cmake_minimum_required(VERSION 3.16)
```
SoapGentoo, line 153: indent everything inside of the input block

eclass/ecm-common.eclass
```diff
@@ -0,0 +1,328 @@
...
+
+	find_package(ECM "5.115.0" REQUIRED NO_MODULE)
+	set(CMAKE_MODULE_PATH \${ECM_MODULE_PATH})
+
+	set(KDE_INSTALL_DOCBUNDLEDIR "${EPREFIX}/usr/share/help" CACHE PATH "")
+
+	include(KDEInstallDirs)
+	include(ECMOptionalAddSubdirectory) # commonly used
+	include(FeatureSummary)
+
```
SoapGentoo, line 164:
```suggestion
```

SoapGentoo (changes requested):

eclass/ecm-common.eclass
```diff
@@ -0,0 +1,347 @@
...
+_ecm-common_docs() {
+	( in_iuse handbook && use handbook ) || return
+	local i
+
+	cat >> CMakeLists.txt <<- _EOF_ || die
+		find_package(KF\${KFSLOT}DocTools REQUIRED)
+		kdoctools_install(po)
+	_EOF_
+
+	if [[ ${ECM_HANDBOOK_DIRS} ]]; then
```
thesamesam, line 214: I don't think we need this condition. We could early-return if it's false - which would look nicer - but the nested for/if should be enough anyway.

eclass/ecm-common.eclass
```diff
@@ -0,0 +1,347 @@
...
+# Override this to add more KF6 has_version checks to pkg_setup(),
+# corresponding with any additional KF6_BDEPEND defined pre-inherit.
+# If false, we'll assume KF5 dependencies are fulfilled via BDEPEND.
+ecm-common-check_deps() {
+	return 1;
+}
+
+ecm-common_pkg_setup() {
+	if has_version -b "kde-frameworks/ki18n:6" &&
+		( in_iuse !handbook || ( in_iuse handbook && use handbook && has_version -b "kde-frameworks/kdoctools:6" ) ) &&
```
thesamesam, line 311: Jeez!

Maybe split into nested functions (although they don't really exist in bash) within `ecm-common_pkg_setup`, so the condition reads nicer.

kde-frameworks/purpose-kaccounts-services/purpose-kaccounts-services-9999.ebuild
```diff
@@ -0,0 +1,41 @@
...
+	!<kde-frameworks/purpose-5.116.0-r2:5
+	!<kde-frameworks/purpose-6.5.0-r1:6
+"
+
+ecm-common-check_deps() {
+	return $(has_version -b "kde-apps/kaccounts-integration:6")
+}
+
+ecm-common_inject_heredoc() {
+	cat >> CMakeLists.txt <<- _EOF_ || die
```
thesamesam, line 31: I actually worry a bit about new services being added here which we miss. Consider a sanity check? (Do find/glob and compare with a hardcoded list in the ebuild?)

kde-frameworks/purpose/purpose-5.239.9999.ebuild
```diff
@@ -62,6 +65,14 @@ src_configure() {
 	ecm_src_configure
 }
 
+src_install() {
+	# Shipped by kde-frameworks/purpose-kaccounts-services package for shared use w/ SLOT 5
+	use kaccounts && ECM_REMOVE_FROM_INSTALL=(
```
thesamesam, line 70: It's indeed much better.

eclass/ecm-common.eclass
```diff
@@ -0,0 +1,355 @@
...
+		die "Value ${ECM_HANDBOOK} is not supported"
+		;;
+esac
+
+if [[ ${ECM_KCM_TARGETS} ]]; then
+	KF5_BDEPEND+=( "kde-frameworks/kcmutils:5" )
+	KF6_BDEPEND+=( "kde-frameworks/kcmutils:6" )
+fi
+
+if [[ ${KF6_BDEPEND} && ${KF5_BDEPEND} ]]; then
```
thesamesam, line 156: We could tidy this up with an array of vars then `${!X}` but I don't think it's worth it for this.

eclass/ecm-common.eclass
```diff
@@ -0,0 +1,355 @@
...
+	_EOF_
+}
+
+# @FUNCTION: ecm-common-check_deps
+# @DESCRIPTION:
+# Override this to add more KF6 has_version checks to pkg_setup(),
+# corresponding with any additional KF6_BDEPEND defined pre-inherit.
+# If false, we'll assume KF5 dependencies are fulfilled via BDEPEND.
+ecm-common-check_deps() {
+	return 1;
```
thesamesam, line 306:
```suggestion
	return 1
```

kde-apps/kate-lib/kate-lib-9999.ebuild
```diff
@@ -39,15 +39,12 @@ DEPEND="
...
+	>=kde-apps/kate-common-${PV}
 "
 
 src_prepare() {
 	ecm_src_prepare
-
-	# delete colliding kate/kwrite translations
-	find po -type f -name "*po" -and ! -name 'kate.po' -delete || die
-	rm -rf po/*/docs || die
+	ecm_punt_po_install
```
thesamesam, line 47: Sort of wonder if we should have `ecm_src_prepare` take a var for it but it's not a big win and it's just shifting the logic. But I generally like being declarative...

Anyway, don't worry about it for now.

thesamesam (approved):

kde-frameworks/purpose-kaccounts-services/purpose-kaccounts-services-9999.ebuild
a17r: We could in this case do sth like `find $S -name "*service.in" | wc -l` and compare that with our currently known hardcoded value.

thesamesam: That would be good enough.

kde-apps/kate-lib/kate-lib-9999.ebuild
a17r: I thought about that as well and even spun it as far as considering separate `${PN}-i18n` and `${PN}-handbook` packages to be referred in aptly named meta vars (then we could trigger such functions by eclass as well), but then with a `-handbook` only package we can not have it as optional RDEPEND again...

eclass/ecm-common.eclass
a17r: Want to come up with a suggestion? I'm struggling there to put more beauty into it for a hopefully only shortlived time where we have KF5-based alternatives even available.

kde-apps/kate-lib/kate-lib-9999.ebuild
thesamesam: Gah. Yet another reason to bump it up my list..

kde-frameworks/purpose-kaccounts-services/purpose-kaccounts-services-9999.ebuild
```diff
@@ -0,0 +1,55 @@
...
+	_EOF_
+}
+
+src_prepare() {
+	ecm-common_src_prepare
+
+	# Safety measure in case new services are added in the future
+	local known_num_of_services=2
+	local found_num_of_services=$(find . -iname "*service.in" | wc -l)
+	if [[ $found_num_of_services != $known_num_of_services ]]; then
```
thesamesam, line 49:
```suggestion
	if [[ ${found_num_of_services} != ${known_num_of_services} ]]; then
```
and so on

eclass/ecm-common.eclass
thesamesam: Inside `ecm-common_pkg_setup`:
```
_handbook_kf6() {
	return in_iuse handbook && use handbook && has_version -b "kde-frameworks/kdoctools:6"
}
```

and so on, then use that in the `if`

eclass/ecm.eclass
```diff
@@ -674,7 +707,7 @@ ecm_src_install() {
 		fi
 	}
 
-	if [[ -n ${_KDE_ORG_ECLASS} && -d "${ED}"/usr/share/metainfo/ ]]; then
+	if [[ -n ${_KDE_ORG_ECLASS} ]] && [[ -d "${ED}"/usr/share/metainfo/ ]]; then
```
thesamesam, line 710: The old one should be a tiny bit faster and is more idiomatic bash (but combining like that isn't valid for POSIX `[`).

eclass/ecm.eclass
```diff
@@ -694,6 +727,14 @@ ecm_src_install() {
...
 		fi
 	fi
+
+	for f in "${ECM_REMOVE_FROM_INSTALL[@]}"; do
+		if [[ -d ${f} ]]; then
+			rm -r "${ED}"${f} || die
+		else
+			rm "${ED}"${f} || die
+		fi
+	done
```
ionenwks, line 737: `[[ -d ]]` check feels like unnecessary complexity, could -r in either case assuming I'm not missing something.

Given `${f}` is unquoted, I assume that supporting globs in `${f}` is wanted, right? *If* didn't want that, could remove the loop too, e.g.:
```bash
	if (( ${#ECM_REMOVE_FROM_INSTALL[@]} )); then
		rm -r "${ECM_REMOVE_FROM_INSTALL[@]/#/${ED}}" || die
	fi
```
But globs sound like they could be useful, so just saying.

ionenwks (commented):
I don't keep up with kde packaging and the problems it has to deal with much, so I have little to say here. Still looked through the code to see if I can pickup anything, but looks good so I couldn't find much to say :)

eclass/ecm.eclass
ionenwks: (albeit fwiw globs may be error prone if someone forgets to quote them and it pickup system files inside the array)

a17r: @sam wdyt? If in doubt probably restrict globbing, otoh this eclass will see very specialised use...

eclass/ecm-common.eclass
a17r: tidied up now with @ionenwks's suggestion, hope that's good enough

eclass/ecm.eclass
thesamesam: I'm inclined to say we allow globbing given the already interesting use of the escaped vars we pass in..

eclass/ecm.eclass
a17r: so ... I'll leave it as it is right now?

thesamesam: I'd prefer it merged together but it's not worth you rebasing.

eclass/ecm-common.eclass
```diff
@@ -0,0 +1,359 @@
...
+	debug-print-function ${FUNCNAME} "$@"
+}
+
+# @FUNCTION: _ecm-common_summary
+# @INTERNAL
+# @DESCRIPTION:
+# Just calls feature_summary
+_ecm-common_summary() {
+	cat >> CMakeLists.txt <<- _EOF_ || die
+
```
SoapGentoo, line 293:
```suggestion
```

SoapGentoo (changes requested):

PR #1002, opened 2024-08-25
ecm-common.eclass: Fix pkg_setup()

a17r:
Ideas for verboseness?

### DIFF
--- a/eclass/ecm-common.eclass
+++ b/eclass/ecm-common.eclass
@@ -309,6 +309,38 @@ _ecm-common_summary() {
 	_EOF_
 }
 
+# @FUNCTION: _ecm-common-check_deps
+# @INTERNAL
+# @DESCRIPTION:
+# Check existence of requested KF6 dependencies.
+_ecm-common-check_deps() {
+	local chk=0
+	case ${1} in
+		i18n)
+			if [[ ${ECM_I18N} ]]; then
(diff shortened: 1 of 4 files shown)


thesamesam (approved):

eclass/ecm-common.eclass
```diff
@@ -309,6 +309,38 @@ _ecm-common_summary() {
 	_EOF_
 }
 
+# @FUNCTION: _ecm-common-check_deps
+# @INTERNAL
+# @DESCRIPTION:
+# Check existence of requested KF6 dependencies.
+_ecm-common-check_deps() {
+	local chk=0
```
thesamesam, line 317: We can do something like:
```
ebegin "Checking for dep type ${1}"
...
eend ${chk}
return ${chk}
```

I don't know if more verbosity beyond that is needed.

thesamesam: What about..
```
ebegin "Checking for dep type ${1} with ECM_I18N=${ECM_I18N}, ECM_HANDBOOK=..."
...
eend ${chk}
return ${chk}
```

Or maybe:
```
ebegin "Checking for dep type ${1}"
...
einfo " ECM_I18N=..." # before each has_version
...
eend ${chk}
return ${chk}
```

thesamesam: Note that `python_has_version` may provide inspiration but may be overkill.

PR #1004, opened 2024-09-02
kde-plasma/kdesu-gui: new package, split from kde-plasma/kde-cli-tools (also new: kde-plasma/kde-cli-tools-common)

a17r:
Estimated ::gentoo revdeps benefitting from less dependencies:
- `net-firewall/pglinux`
- `sys-block/gparted`
- `x11-misc/spacefm`
- ... maybe some that failed to declare the USE dep.

Need to check yet, but *maybe* we could stabilise the resulting smaller depgraph for x86 then to be able to drop Plasma 5.

Furthermore:
`app-office/libreoffice` (and apparently also `x11-misc/libinput-gestures`) are optionally using kioclient from kde-cli-tools. Maybe can be dropped for xdg-utils only (but only in the long term)?


kde-plasma/kdesu-gui/kdesu-gui-6.1.49.9999.ebuild
```diff
@@ -0,0 +1,53 @@
...
+src_configure() {
+	local mycmakeargs=(
+		-DWITH_X11=$(usex X)
+	)
+	ecm_src_configure
+}
+
+src_install() {
+	ecm_src_install
+	dosym ../libexec/kf6/kdesu /usr/bin/kdesu
```
thesamesam, line 52: Just dosym -r? This is an EAPI 8 world, baby

thesamesam (approved):

PR #1006, opened 2024-10-05
media-gfx/kgeotag: Adjust case of CHANGELOG.rst

SigHunter:
Upstream [changed](https://invent.kde.org/graphics/kgeotag/-/commit/91246a200c3338b60a377ac94abafcd5a3aae98e) the case of ChangeLog.rst to CHANGELOG.rst and the ebuild failed when installing the DOCS with 
`FileNotFoundError: [Errno 2] No such file or directory: b'ChangeLog.rst'`

media-gfx/kgeotag-1.5.0::gentoo already has this, the 9999::kde was missing it.

### DIFF
--- a/media-gfx/kgeotag/kgeotag-9999.ebuild
+++ b/media-gfx/kgeotag/kgeotag-9999.ebuild
@@ -34,4 +34,4 @@ DEPEND="
 "
 RDEPEND="${DEPEND}"
 
-DOCS=( ChangeLog.rst README.md )
+DOCS=( CHANGELOG.rst README.md )


SigHunter:
oh, 1.6.0 is now here too, shall I bring the 9999 in line with the 1.6.0::gentoo? (KFMIN, QTMIN and the cmake patch)

a17r:
It is the other way round wrt dependencies - 9999 is template for version bumps, so please drop the KFMIN/QTMIN syncing.

Patching 9999 ebuilds is generally a bad idea and it probably wasn't done in the expectation of quickly upstream that patch. Submitted now: https://invent.kde.org/graphics/kgeotag/-/merge_requests/19

a17r (changes requested):
Thanks for the DOCS fix, will merge as soon as reduced to that.

a17r (approved):

PR #1010, opened 2024-11-12
app-misc/francis: new package, add 9999

parona-source:
Not super sure on the category.

https://apps.kde.org/francis/

### DIFF
--- a/app-misc/francis/francis-9999.ebuild
+++ b/app-misc/francis/francis-9999.ebuild
@@ -1,0 +1,29 @@
+# Copyright 2024 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+KDE_ORG_CATEGORY="utilities"
+KFMIN=6.5.0
+QTMIN=6.7.2
+inherit ecm gear.kde.org
+
+DESCRIPTION="Productivity application using the well-known pomodoro technique"
+HOMEPAGE="https://apps.kde.org/francis/"
+
+LICENSE="BSD BSD-2 CC0-1.0 GPL-3+ LGPL-2+ LGPL-2.1+"
+SLOT="0"
+KEYWORDS=""
+
+DEPEND="
+	>=dev-libs/kirigami-addons-1.0:6
+	>=dev-qt/qtbase-${QTMIN}:6[dbus,gui,widgets]
+	>=dev-qt/qtdeclarative-${QTMIN}:6[svg,widgets]
+	>=dev-qt/qtsvg-${QTMIN}:6
+	>=kde-frameworks/kconfig-${KFMIN}:6
+	>=kde-frameworks/kcoreaddons-${KFMIN}:6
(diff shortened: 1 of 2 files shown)


app-misc/francis/francis-9999.ebuild
```diff
@@ -0,0 +1,29 @@
...
+
+EAPI=8
+
+KDE_ORG_CATEGORY="utilities"
+KFMIN=6.5.0
+QTMIN=6.6.0
+inherit ecm gear.kde.org
+
+DESCRIPTION="Francis uses the well-known Pomodoro technique to help you get more productive."
+HOMEPAGE="https://invent.kde.org/utilities/francis/"
```
a17r, line 12: Use `https://apps.kde.org/francis/` instead

app-misc/francis/francis-9999.ebuild
```diff
@@ -0,0 +1,29 @@
+# Copyright 2024 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+KDE_ORG_CATEGORY="utilities"
+KFMIN=6.5.0
+QTMIN=6.6.0
```
a17r, line 8: We use `6.7.2` as baseline in all other KDE Gear packages, so this should too. The reason for that was yet another ABI break on that release so it was necessary to enforce this to avoid breakages.

app-misc/francis/francis-9999.ebuild
```diff
@@ -0,0 +1,29 @@
...
+inherit ecm gear.kde.org
+
+DESCRIPTION="Francis uses the well-known Pomodoro technique to help you get more productive."
+HOMEPAGE="https://invent.kde.org/utilities/francis/"
+
+LICENSE="BSD BSD-2 CC0-1.0 GPL-3+ LGPL-2+ LGPL-2.1+"
+SLOT="6"
```
a17r, line 15: It's not a library or KIO worker, so just use SLOT 0 there.

app-misc/francis/francis-9999.ebuild
```diff
@@ -0,0 +1,29 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+KDE_ORG_CATEGORY="utilities"
+KFMIN=6.5.0
+QTMIN=6.6.0
+inherit ecm gear.kde.org
+
+DESCRIPTION="Francis uses the well-known Pomodoro technique to help you get more productive."
```
a17r, line 11:
```suggestion
DESCRIPTION="Productivity application using the well-known pomodoro technique"
```

`${PN}` shall never be part of `DESCRIPTION`, and we don't write full sentences there either, so no full stop either.

a17r (changes requested):

a17r (approved):
> Not super sure on the category.

There's x11-misc/flow-pomodoro as precedent, so app-misc sounds about right. Thanks!

PR #1013, opened 2024-12-06
*.kde.org: Push PVCUT from ebuilds into new eclass var

a17r:
Currently an idea, open to suggestions.

Since meta variables are unlikely to get any less I thought it would make sense to try and reduce where sensical.

### DIFF
--- a/eclass/frameworks.kde.org.eclass
+++ b/eclass/frameworks.kde.org.eclass
@@ -25,6 +25,13 @@ esac
 if [[ -z ${_FRAMEWORKS_KDE_ORG_ECLASS} ]]; then
 _FRAMEWORKS_KDE_ORG_ECLASS=1
 
+# @ECLASS_VARIABLE: KDE_CATV
+# @DESCRIPTION:
+# Holds main Frameworks release number (major.minor) for use on same-category
+# dependencies.
+KDE_CATV=$(ver_cut 1-2)
+readonly KDE_CATV
+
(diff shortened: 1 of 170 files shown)


eclass/frameworks.kde.org.eclass
```diff
@@ -25,6 +25,13 @@ esac
 if [[ -z ${_FRAMEWORKS_KDE_ORG_ECLASS} ]]; then
 _FRAMEWORKS_KDE_ORG_ECLASS=1
 
+# @ECLASS_VARIABLE: KDE_CATV
+# @DESCRIPTION:
+# Holds main Frameworks release number (major.minor) for use on same-category
+# dependencies.
+KDE_CATV=$(ver_cut 1-2)
```
thesamesam, line 32: Would KF_CATV or KDE_KF_KCATV not be better?

a17r: I chose `KDE_` as the informal `*.kde.org` eclass prefix. It should be just as generic as `KFMIN` and `QTMIN` already are, so there is no unnecessary potential for mixups.

thesamesam: Oh, ok.

thesamesam: Nah, I understand what you were going for now -- the eclass is solely for these packages, we might use this same variable name in other such cat-specific eclasses too. It doesn't make sense to allow it to be mixed or anything.

thesamesam (approved):

PR #1015, opened 2025-01-17
dev-libs/kopeninghours: use Python cmake module instead of patching

parona-source:
Upstream commit: c73771a4e5cc1327f65092206889b995d2200ef7


parona-source:
Im currently doing the ecm eclass changes in another branch, this came up while mass building. So ill look at ecm_pkg_setup there.

PR #1030, opened 2025-02-13
kde-plasma/plasma-meta: depend on plasma-desktop[input_devices_wacom] for wacom USE

ArsenArsen:
This USE flag is required to install the Drawing Tablet KCM, which is very useful for someone with a drawing tablet.


a17r (changes requested):

a17r (changes requested):
missing sign-off

PR #1032, opened 2025-04-14
kde-apps/ark: 7zip[symlink] as an alternative 

parona-source:
In the future its planned to use libarchive for 7z archives. In my tests ark (ark 24.12.3) worked correctly (opening and creating 7z archive with or without passwords) when I had 7zip[symlink] installed and p7zip in package.provided.

https://invent.kde.org/utilities/ark/-/merge_requests/238
https://bugs.kde.org/show_bug.cgi?id=458334
https://bugs.kde.org/show_bug.cgi?id=468240


thesamesam:
Thanks! For future reference, I'd include the testing info in the commit message too.

PR #1033, opened 2025-04-18
kde-apps/calendarjanitor: update loggingcategory patch

parona-source:
Upstream-Commit: 0100ee00401aa5a2752aacc41f628a29f333efe5

### DIFF
--- a/kde-apps/calendarjanitor/calendarjanitor-9999.ebuild
+++ b/kde-apps/calendarjanitor/calendarjanitor-9999.ebuild
@@ -31,7 +31,7 @@ RDEPEND="${DEPEND}
 	>=kde-apps/akonadi-calendar-tools-common-${PV}
 "
 
-PATCHES=( "${FILESDIR}/${PN}-24.05.2-loggingcategory.patch" )
+PATCHES=( "${FILESDIR}/${PN}-25.07.70-loggingcategory.patch" )
 
 src_prepare() {
 	ecm_src_prepare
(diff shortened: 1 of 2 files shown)


kde-apps/calendarjanitor/calendarjanitor-9999.ebuild
```diff
@@ -31,7 +31,7 @@ RDEPEND="${DEPEND}
 	>=kde-apps/akonadi-calendar-tools-common-${PV}
 "
 
-PATCHES=( "${FILESDIR}/${PN}-24.05.2-loggingcategory.patch" )
+PATCHES=( "${FILESDIR}/${PN}-9999-loggingcategory.patch" )
```
a17r, line 34:
```suggestion
PATCHES=( "${FILESDIR}/${PN}-25.07.70-loggingcategory.patch" )
```

Use the currently set upstream version in CMakeLists.txt. I don't want to have to rename it to make sense when the next Gear version is released.

a17r (changes requested):

kde-apps/calendarjanitor/calendarjanitor-9999.ebuild
parona-source: I'm unsure where to find that in this case. I can only find the PIM_VERSION which is used for the project and is a non Gear version.

https://invent.kde.org/pim/akonadi-calendar-tools/

a17r: Yeah, it's unfortunate in this case where PIM uses its own internal one. But we know the schedule.

a17r (approved):

PR #1034, opened 2025-04-23
net-im/tokodon: set KFMIN=6.11.0

CyberTailor:
Tokodon 25.04 fails to configure if an older version of frameworks is
installed.


a17r (commented):
What about 9999?

PR #1035, opened 2025-04-24
kde-apps/lokalize: remove python scripts and re-enable tests

parona-source:


a17r:
Thanks, this  is coming just at the right time to unblock py3.13 stabilisation. Thankfully, I didn't put any more work into it other than figuring out dependencies.

You could reference https://bugs.gentoo.org/952507 from this commit and we'll backport to 24.12 and 25.04 afterwards.

a17r:
Yes please, and thanks in advance!

kde-apps/lokalize/lokalize-25.04.49.9999.ebuild
```diff
@@ -41,16 +39,12 @@ DEPEND="${PYTHON_DEPS}
...
-	')
-"
+RDEPEND="${DEPEND}"
 
-src_install() {
-	ecm_src_install
-	rm "${ED}"/usr/share/lokalize/scripts/msgmerge.{py,rc} || die
-	python_fix_shebang "${ED}"/usr/share/${PN}
+src_prepare() {
+	cmake_src_prepare
```
a17r, line 45: we're using ecm.eclass, so it must be calling `ecm_src_prepare` instead.

a17r (changes requested):
What about tests for 25.04?

a17r (approved):
Thanks!

PR #1039, opened 2025-05-03
Add subslot operators for future libxml2 subslot bump

parona-source:
https://github.com/gentoo/gentoo/pull/41919


a17r:
Where is that incoming libxml2 subslot bump?

parona-source:
> Where is that incoming libxml2 subslot bump?

https://bugs.gentoo.org/954318

Due to how large the PR they actual libxml2 bump isn't shown first.

PR #1040, opened 2025-05-09
kde-frameworks/extra-cmake-modules: include implicit include

parona-source:
kinfocenter notably relies on this implicit include that this patch would otherwise remove.

Depending on the release order, this may have to be backported so that kinfocenter's next release works.

### DIFF
--- a/kde-frameworks/extra-cmake-modules/extra-cmake-modules-9999.ebuild
+++ b/kde-frameworks/extra-cmake-modules/extra-cmake-modules-9999.ebuild
@@ -38,7 +38,7 @@ BDEPEND="
 PATCHES=(
 	"${FILESDIR}/${PN}-5.49.0-no-fatal-warnings.patch"
 	"${FILESDIR}/${PN}-5.93.0-skip-ecm_add_test-early.patch"
-	"${FILESDIR}/${PN}-5.245.0-disable-qmlplugindump.patch"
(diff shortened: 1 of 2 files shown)


a17r:
@parona-source @thesamesam Why aren't we fixing implicit includes in consumers instead? Seems like it is possible: https://invent.kde.org/plasma/kinfocenter/-/issues/6

parona-source:
> @parona-source @thesamesam Why aren't we fixing implicit includes in consumers instead? Seems like it is possible: https://invent.kde.org/plasma/kinfocenter/-/issues/6 https://invent.kde.org/plasma/kinfocenter/-/commit/b2fd0e4cb249980c20910038bde21bc676529c7c

Really a matter of where the line is with acceptable breakage from downstream patches that gets sent upstream.

a17r:
An implicit dependency is a bug no matter what. We are the ones who find it, and we can fix it, and upstream evidently accept these fixes.

thesamesam:
I don't disagree that we should fix those, but I don't want upstream to be confused by well-meaning users reporting issues there that got introduced or exposed by a downstream patch without us being there to explain context.

a17r:
This patch has existed for quite some time and so far, kinfocenter was the only case, swiftly fixed by upstream after reporting. I mean, I should probably go and try upstream it some time.

thesamesam:
OK, I'll revert it then.

PR #1042, opened 2025-05-17
drkonqi legacy

a17r:


thesamesam (changes requested):
Needs rationale in commit message. I'm also really not a fan of slots which block each other (they're then not slots, and it's poor UX for users). But wait, there's no blocker here either?

a17r:
well, it's a draft ;)

if we do not slot it, then the only alternative would be different `PN` imo, since as we know "blocked" upgrades are also bad UX.

thesamesam:
Just mask the older versions outside of features/systemd.

PR #1043, opened 2025-05-17
CMake 4: detection of incompatible CMakeList.txt

a17r:


eclass/cmake.eclass
```diff
@@ -243,14 +243,44 @@ _cmake_modify-cmakelists() {
 	# Only edit the files once
 	grep -qs "<<< Gentoo configuration >>>" "${CMAKE_USE_DIR}"/CMakeLists.txt && return 0
 
+	local x re="VERSION( .*\.\.\.| )(([[:digit:]]+)\.([[:digit:]]+))"
```
thesamesam, line 246: Can we do what we discussed and add the magic compat arg accordingly when we hit this case? i.e. QA notice + make it work?

thesamesam (approved):
Seems okay with some opt-out var as discussed on ML? I don't care if it's fine-grained (excluding subdirs) or just some generic CMAKE_NO_VERSION_WARN or whatever.

eclass/cmake.eclass
```diff
@@ -362,6 +362,17 @@ cmake_src_configure() {
 	# Fix xdg collision with sandbox
 	xdg_environment_reset
 
+	local x re="cmake_minimum_required *\( *VERSION( .*\.\.\.| )(([[:digit:]]+)\.([[:digit:]]+))"
+	local ver cmreq_isold
+	for x in $(find "${CMAKE_USE_DIR}" -type f -iname "CMakeLists.txt" -maxdepth 2); do
```
thesamesam, line 367: Let's change this to..
```
        while read -d '' -r file ; do
                ...
        done < <(find ... -print0)
```
so we handle space dirs correctly.

eclass/cmake.eclass
```diff
@@ -362,6 +368,17 @@ cmake_src_configure() {
 	# Fix xdg collision with sandbox
 	xdg_environment_reset
 
+	local file ver cmreq_isold
+	if ! [[ ${CMAKE_QA_COMPAT_SKIP} ]]; then
+		while read -d '' -r file ; do
+			ver=$(sed -ne "/cmake_minimum_required/ s/.*\(\.\.\.\|\s\)\([0-9.]*\)\s.*$/\2/p" ${file} 2>/dev/null)
```
mgorny, line 374:
```suggestion
			ver=$(sed -ne "/cmake_minimum_required/ s/.*\(\.\.\.\|\s\)\([0-9.]*\)\s.*$/\2/p" "${file}" 2>/dev/null)
```

You may also want `{p;q}` for the same behavior, I think.

a17r:
... and already had to fix the `sed` after some testing. pushing to kde overlay now, please give it a spin.

PR #1047, opened 2025-06-13
cmake.eclass: Add src_install() check for CMake-4-incompatible modules

a17r:


eclass/cmake.eclass
```diff
@@ -698,6 +716,25 @@ cmake_src_install() {
...
 		popd > /dev/null || die
 	fi
+
+	local file files=()
+	while read -d '' -r file ; do
+		# Detect unsupported minimum CMake versions unless CMAKE_QA_COMPAT_SKIP is set
+		if ! [[ ${CMAKE_QA_COMPAT_SKIP} ]]; then
+			_cmake_minreqver-detect "3.5" "${file}" || files+=( "${file}" )
+		fi
+	done < <(find "${D}" -type f -iname "*.cmake" -print0)
```
mgorny, line 726: I would:
```suggestion
	done < <(find "${D}" -type f -iname "*.cmake" -print0 || die)
```

eclass/cmake.eclass
```diff
@@ -698,6 +716,25 @@ cmake_src_install() {
...
+	if [[ ${#files[*]} -gt 0 ]]; then
+		eqawarn "QA Notice: Package installs CMake module(s) incompatible with CMake 4,"
+		eqawarn "breaking any packages relying on it:"
+		eqawarn
+		for file in "${files[@]}"; do
+			eqawarn "    ${file/${D}/}"
```
mgorny, line 732: Wouldn't:

```suggestion
			eqawarn "    ${file#${D}}"
```

work?

eclass/cmake.eclass
```diff
@@ -123,6 +123,11 @@ fi
 # If set, skip detection of CMakeLists.txt unsupported in CMake 4 in case of
 # false positives (e.g. unused outdated bundled libs).
 
+# @ECLASS_VARIABLE: _CMAKE_MINREQVER_UNSUPPORTED
+# @DEFAULT_UNSET
+# @DESCRIPTION:
+# Set to true if an unsupported cmake_minimum_required value was detected.
```
mgorny, line 129: Isn't this `@INTERNAL`? Right now it sounds like the user is supposed to set it.

eclass/cmake.eclass
```diff
@@ -238,6 +243,25 @@ _cmake_check_build_dir() {
...
 }
 
+# @FUNCTION: _cmake_minreqver-detect
+# @USAGE: <lt-version> <path>
+# @INTERNAL
+# @DESCRIPTION:
+# Internal function for detecting occurrence of lower-than-specified
+# <lt-version> in cmake_minimum_required of a given CMake file <path>.
+# Returns 1 if the regex matched (a lower-than-specified version found).
+_cmake_minreqver-detect() {
```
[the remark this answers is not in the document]
a17r: @mgorny we don't ever need to look in the other direction, so `_cmake_minreqver_lt` would be fine.

and then reverse the return value logic, so success follows the wording of the function name, I guess

eclass/cmake.eclass
ionenwks: Also, quotes for whomever had the funny ideas to have globs in their dir names
```suggestion
			eqawarn "    ${file#"${D}"}"
```
As a minor optimization, just `${file}` here and do `files+=( "${file#"${D}"}" )` earlier on not to store ${D} for nothing.

eclass/cmake.eclass
```diff
@@ -238,6 +244,25 @@ _cmake_check_build_dir() {
...
+	local ver chk=0
+	ver=$(sed -ne "/cmake_minimum_required/I{s/.*\(\.\.\.*\|\s\)\([0-9][0-9.]*\)\([)]\|\s\).*$/\2/p;q}" \
+		"${2}" 2>/dev/null \
+	)
+	if [[ -n $ver ]] && ver_test $ver -lt "${1}"; then
```
ionenwks, line 259: ${1} gets to have braces but $ver doesn't? Poor thing.

thesamesam (approved):

PR #1049, opened 2025-06-17
dev-qt/qtwayland -> dev-qt/qtbase[wayland] depchange

a17r:


thesamesam:
Can you populate the commit messages?

a17r:
I'd prefer not to, this is already costing enough time

eli-schwartz:
Based on #gentoo-qt logs and spelunking "probable cause" via qtbase commit logs, this is presumably:

> The parts of qtwayland we depend on have been moved into `qtbase[wayland]` for Qt 6.10, and for Qt 6.9 it's technically correct because it PDEPENDs on qtwayland.

This is an outright PMS violation, so it is not "technically correct".
https://dev.gentoo.org/~ulm/pms/head/pms.html#section-8.1
- Runtime dependencies (RDEPEND). These must be installed and usable before the results of an ebuild merging are treated as usable.
- Post dependencies (PDEPEND). These must be installed at some point before the package manager finishes the batch of installs.

(yes, pdep is useless 99% of the time)

The distinction of the latter is *precisely* that it need not be installed before the results of the package are treated as usable. It is "if installing X, behave as though Y was also added to world / the CLI". If portage needs to merge `qtbase -> randompackage -> qtwayland`, it can (and @thesamesam points out due to https://bugs.gentoo.org/180045 there is best-effort code to work around this absent cycles, but still).

Frankly I don't see the point. Leaving it alone will still recursively depend on all the right things, plus an extra package that might otherwise be a depclean candidate? Why not simply leave this alone until it can be actually-correct, by requiring QTMIN of 6.10

ionenwks:
> Based on #gentoo-qt logs and spelunking "probable cause" via qtbase commit logs, this is presumably:
> 
> > The parts of qtwayland we depend on have been moved into `qtbase[wayland]` for Qt 6.10, and for Qt 6.9 it's technically correct because qtbase PDEPENDs on qtwayland.
> 
> This is an outright PMS violation, so it is not "technically correct". https://dev.gentoo.org/~ulm/pms/head/pms.html#section-8.1
> 
> * Runtime dependencies (RDEPEND). These must be installed and usable before the results of an ebuild merging are treated as usable.
> * Post dependencies (PDEPEND). These must be installed at some point before the package manager finishes the batch of installs.

Hadn't noticed this "at some point" bit rather than being something like "earliest possible"... which I find odd given it makes PDEPEND kind of useless. If it's something needed at runtime, and then other packages expect that package work... what then? Of course, can never resolve this for circular deps but you can for other packages.

According to sam, portage will still prioritize installing PDEPEND once it can though -- not to say what other PMs do.

Either way, I'll change the wiki later to not call it correct and maybe update the suggestion depending on what we do here.

> 
> (yes, pdep is useless 99% of the time)
> 
> The distinction of the latter is _precisely_ that it need not be installed before the results of the package are treated as usable. It is "if installing X, behave as though Y was also added to world / the CLI". If portage needs to merge `qtbase -> randompackage -> qtwayland`, it can (and @thesamesam points out due to https://bugs.gentoo.org/180045 there is best-effort code to work around this absent cycles, but still).
> 
> Frankly I don't see the point. Leaving it alone will still recursively depend on all the right things, plus an extra package that might otherwise be a depclean candidate? Why not simply leave this alone until it can be actually-correct, by requiring QTMIN of 6.10

Primarily because `dev-qt/qtwayland-6.10` will now always build the compositor bits (the USE was removed+enabled), and it's a large (wasteful) build -- nearly the only users that will need qtwayland-6.10 are those that have telegram-desktop or mathematica installed.

In most cases we could do `|| ( >=qtbase-6.10[wayland] <qtwayland-6.10  )` but what prompted to avoid this is packages depending on private qtwayland APIs and needing `:=` and doing `|| ( := := )` is utterly broken.

Options I see:
1. let users build a large package they don't need and change deps later, likely when Qt 6.10 is stable, it's not great but it's not like users will have to deal with this forever either
2. cheat and rely on PDEPEND being done early, eventually <6.10 will be gone and it won't be particularly wrong anymore
3. "best effort" use `|| ( )` where possible, and cheat only where `:=` is needed, that may be good enough to avoid (most) ordering issues if they do exist -- can cleanup the alternative in `|| ( )` without a revbump later
4. rename real qtwayland-6.10 and make the old one a meta that pulls qtbase[wayland], but I'm not thrilled at keeping a permanent renaming mismatching upstream over a temporary workaround and saving build time

Personally I'm fine with any of 1-3.

***Either way*** 6.10 won't release till September (or later), and even then we'll probably keep 6.10.0 masked until 6.10.1 so there is no urgency here. I could see this taking until late October or November for ~arch to get Qt 6.10.

eli-schwartz:
> Hadn't noticed this "at some point" bit rather than being something like "earliest possible"... which I find odd given it makes PDEPEND kind of useless. If it's something needed at runtime, and then other packages expect that package work... what then? Of course, can never resolve this for circular deps but you can for other packages.

In terms of PM design, what you want is in fact an rdepend, no more, no less. There is actually no reason rdepends should be prevented from merging in any order -- i.e. pure-rdepend cycles are reasonable and permissible -- except in the specific case where one or the other runs pkg_postinst, in which case the package itself needs to be in "mostly-satisfied" state w.r.t. dependencies as a prerequisite for pkg_postinst, and successfully running pkg_postinst is what causes a transition to "satisfied".

More generally, merging a package consists of several aspects, all of which are optional:
- merge D -> ROOT  (if D exists)
- setup dependencies (R, I, P variants)
- execute shell scriptlet hooks (phase function for `pkg_*`)

and 3) has a dependency on both 1) and 2), but 1) does not directly depend on 2). Sadly, due to https://bugs.gentoo.org/937150 I don't think we're actually allowed to skip 3), ever.

ionenwks:
> Options I see:
> 
> 1. let users build a large package they don't need and change deps later, likely when Qt 6.10 is stable, it's not great but it's not like users will have to deal with this forever either
> 2. cheat and rely on PDEPEND being done early, eventually <6.10 will be gone and it won't be particularly wrong anymore
> 3. "best effort" use `|| ( )` where possible, and cheat only where `:=` is needed, that may be good enough to avoid (most) ordering issues if they do exist -- can cleanup the alternative in `|| ( )` without a revbump later
> 4. rename real qtwayland-6.10 and make the old one a meta that pulls qtbase[wayland], but I'm not thrilled at keeping a permanent renaming mismatching upstream over a temporary workaround and saving build time
> 
> Personally I'm fine with any of 1-3.
> 
> ***Either way*** 6.10 won't release till September (or later, formerly August but they delayed it), and even then we'll probably keep 6.10.0 masked until 6.10.1 so there is no urgency here. I could see this taking until late October or November for ~arch to get Qt 6.10.

@thesamesam any opinion on which we should do? @eli-schwartz too, albeit assume would still prefer 1) per the previous comment.

Note that if risk of real world issues is minimal when using portage (ignoring "violates PMS"), 2) would save work for asturm.

ionenwks:
> Hm, guess can think of a fifth option if want pmspec happiness without needing much work.
> 
> Keep the current changes as-is but add a temporary workaround `|| ( >=qtbase-6.10[wayland] <qtwayland-6.10 )` on top (without :=, that can be kept on the current qtbase dep), the idea is that it does not control the main dependency but is just there to say "I want one of these _first_". Later the line can just be dropped without changing anything else when <6.10 is not much of a concern anymore.

For a sec I thought that this may not work because it may see qtwayland-6.9 and think "this satisfies it", however it really shouldn't given qtwayland-6.9 is in a broken state after qtbase-6.10 is merged (it depends on ~qtbase-6.9) -- portage did sometime mis-order in that situation before though but I think these issues are resolved.

a17r:
> Keep the current changes as-is but add a temporary workaround `|| ( >=qtbase-6.10[wayland] <qtwayland-6.10 )` on top (without :=, that can be kept on the current qtbase dep), the idea is that it does not control the main dependency but is just there to say "I want one of these _first_".

jfyi this is now implemented for all the affected ebuilds.

PR #1065, opened 2025-08-13
Dependency updates and slight random checks for missing or incorrect dependencies

parona-source:
Upstream requiring ffmpeg support for dragon is a notable change.

Addition of kcoreaddons to RDEPEND for the the kio thumbnailer packages is a bit odd if you only look at the cmake. But its clear that it gets linked against the plugins in addition to the usage of its cmake macro.

Updated the untagged project kde-misc/kregexpeditor to use KFMIN/QTMIN just like every other package in the repo.


kde-apps/dragon/dragon-9999.ebuild
```diff
@@ -16,17 +16,21 @@ SLOT="6"
...
-	>=dev-qt/qtmultimedia-${QTMIN}:6
+	>=dev-qt/qtmultimedia-${QTMIN}:6=[ffmpeg]
 	>=kde-frameworks/kconfig-${KFMIN}:6
 	>=kde-frameworks/kcoreaddons-${KFMIN}:6
 	>=kde-frameworks/kcrash-${KFMIN}:6
 	>=kde-frameworks/ki18n-${KFMIN}:6
 	>=kde-frameworks/kio-${KFMIN}:6
 	>=kde-frameworks/kirigami-${KFMIN}:6
 	>=kde-frameworks/kservice-${KFMIN}:6
+	>=kde-frameworks/kwindowsystem-${KFMIN}:6
```
a17r, line 32: `grep -R include.*KX` reveals `#include <KX11Extras>` so it'll also need X USEdep there. (and all older versions)

kde-apps/kio-extras/kio-extras-9999.ebuild
```diff
@@ -60,11 +56,15 @@ DEPEND="
...
 	phonon? ( >=media-libs/phonon-4.12.0[qt6(+)] )
 	samba? (
+		>=kde-frameworks/kdnssd-${KFMIN}:6
 		net-fs/samba[client]
 		>=net-libs/kdsoap-2.2.0:=[qt6(+)]
 		>=net-libs/kdsoap-ws-discovery-client-0.3.0
 	)
-	sftp? ( net-libs/libssh:=[sftp] )
+	sftp? (
+		dev-libs/qcoro
```
a17r, line 65: this should actually be a `DEPEND`-only, can you check?

kde-apps/konqueror/konqueror-25.07.90.ebuild
```diff
@@ -50,11 +50,10 @@ COMMON_DEPEND="
 	>=kde-frameworks/kxmlgui-${KFMIN}:6
 	>=kde-frameworks/sonnet-${KFMIN}:6
 	sys-libs/zlib
-	speech? ( >=dev-qt/qtspeech-${QTMIN}:6 )
-"
-DEPEND="${COMMON_DEPEND}
 	activities? ( kde-plasma/plasma-activities:6 )
+	speech? ( >=dev-qt/qtspeech-${QTMIN}:6 )
 "
+DEPEND="${COMMON_DEPEND}"
```
a17r, line 56: In that case it would seem the `COMMON_DEPEND` is obsolete?

kde-apps/libksieve/libksieve-25.07.90.ebuild
```diff
@@ -21,7 +21,7 @@ IUSE="speech"
 RESTRICT="test"
 
 RDEPEND="
-	dev-libs/cyrus-sasl
+	dev-libs/cyrus-sasl:2
```
a17r, line 24: While that is fine to do, it was also fine not to declare the slot in the absence of any other slot available (for as long as I remember).

a17r (changes requested):
Good stuff, thanks! Will review more when not on the road anymore.

kde-apps/libksieve/libksieve-25.07.90.ebuild
thesamesam: Indeed (though I don't object to adding it either).

kde-misc/kregexpeditor/kregexpeditor-9999.ebuild
```diff
@@ -4,7 +4,8 @@
 EAPI=8
 
 ECM_HANDBOOK="forceoptional"
-KFMIN=6.3.0
+QTMIN=6.7
+KFMIN=6.5.0
 inherit ecm kde.org
```
a17r, line 9:
```suggestion
KFMIN=6.16.0
QTMIN=6.9.1
inherit ecm kde.org xdg
```

- keep it sorted and use current kde overlay [KDE Gear] 9999 defaults
- xdg: this installs a desktop file ^ and necessary to do if KFMIN >=6.9.0

kde-apps/kio-extras/kio-extras-9999.ebuild
a17r: we don't generally keep these in RDEPEND if stripped by --as-needed - cmake `target_link_libraries` is not authoritative methinks @thesamesam

kde-apps/libksieve/libksieve-25.07.90.ebuild
a17r: sure. what matters more is pkgcheck, which is silent here, though.

kde-apps/kio-extras/kio-extras-9999.ebuild
parona-source: My rationale is that this is an implementation detail so it could change later without notice (especially if upstream documentation instructs to link it). So I wouldn't want to look at it again for the off chance if gets changed in qcoro.

thesamesam: We should include it if it's documented especially, because it could be that some e.g. new template gets instantiated and suddenly it is needed.

a17r: I've basically not `RDEPEND`ed on `dev-libs/qcoro` anywhere yet. Could you create a separate PR to deal with all affected revdeps?

a17r (approved):

a17r:
> Addition of kcoreaddons to RDEPEND for the the kio thumbnailer packages is a bit odd if you only look at the cmake. But its clear that it gets linked against the plugins in addition to the usage of its cmake macro.

That should be sent upstream I think.

a17r (approved):
(excellent work, thanks!)

PR #1067, opened 2025-08-16
kde-apps/{kate,kate-addons,kwrite}: new kate-lib files

parona-source:
Upstream commit: c2aa4f788764d6755076d6ba950a014bc5c8b3cd
Upstream commit: 98aab64cf96ed8903b2c58d4b1c5257279c7af42

### DIFF
--- a/kde-apps/kate-addons/kate-addons-9999.ebuild
+++ b/kde-apps/kate-addons/kate-addons-9999.ebuild
@@ -89,6 +89,8 @@ src_install() {
 
 	# provided by kde-apps/kate-lib
 	rm -v "${ED}"/usr/$(get_libdir)/libkateprivate.so.* || die
+	rm -v "${ED}"/usr/$(get_libdir)/qt6/plugins/kf6/kio/kio_kateexec.so || die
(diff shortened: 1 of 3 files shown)


kde-apps/kate-addons/kate-addons-9999.ebuild
```diff
@@ -89,6 +89,8 @@ src_install() {
 
 	# provided by kde-apps/kate-lib
 	rm -v "${ED}"/usr/$(get_libdir)/libkateprivate.so.* || die
+	rm -v "${ED}"/usr/$(get_libdir)/qt6/plugins/kf6/kio/kio_kateexec.so || die
```
a17r, line 92: Is that needed at build-time? I doubt it. Could you try disable it already at cmake stage, so it isn't built in the first place except for kate-lib?

a17r (changes requested):

kde-apps/kate-addons/kate-addons-9999.ebuild
parona-source: Yeah, this library in particular could be skipped easily.

PR #1071, opened 2025-09-25
kde-apps/audiocd-kio: patch redundant

parona-source:
Upstream commit: 7e6c8e08c919216b8d456a8c2fb0b4c9484f5d4c

The patch was broken either way as it wasn't updated for kf6.

### DIFF
--- a/kde-apps/audiocd-kio/audiocd-kio-25.08.49.9999.ebuild
+++ b/kde-apps/audiocd-kio/audiocd-kio-25.08.49.9999.ebuild
@@ -37,7 +37,7 @@ DEPEND="
 "
 RDEPEND="${DEPEND}"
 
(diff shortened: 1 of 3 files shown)


kde-apps/audiocd-kio/audiocd-kio-25.08.49.9999.ebuild
```diff
@@ -37,7 +37,7 @@ DEPEND="
 "
 RDEPEND="${DEPEND}"
 
-PATCHES=( "${FILESDIR}/${PN}-19.04.0-handbook.patch" )
+PATCHES=( "${FILESDIR}/${PN}-23.03.80-handbook.patch" )
```
a17r, line 40: It should be package version introduced patch with; you're not going to retroactively fix old and cleaned up versions with that? ;)

a17r (changes requested):

PR #1072, opened 2025-09-28
cmake.eclass improvements

a17r:
- cmake.eclass: Reorder arguments in cmake_src_install cmake_build call
- cmake.eclass: allow nonfatal cmake_src_test
- cmake.eclass: cmake.eclass: Provide cmake_prepare() to solve conflicts w/ other eclass (>=EAPI-8)
- cmake.eclass: fix pms violation

### DIFF
--- a/eclass/cmake.eclass
+++ b/eclass/cmake.eclass
@@ -730,7 +730,9 @@ cmake_src_test() {
 			eerror "--END TEST LOG----------------------------------------------------------------"
 			die -n "Tests failed."
 		else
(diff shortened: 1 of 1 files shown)


thesamesam (commented):
The "do not call ... unconditionally" commit does still call `default_src_prepare` unconditionally for non-EAPI 7, right? That's OK for now as it's good prep work for a future change, but it doesn't make the change the commit message says.

thesamesam (approved):

PR #1073, opened 2025-10-01
cmake.eclass: Really warn all EAPI-7 consumers that support will be dropped

a17r:
... after 2025-11-01.


eclass/cmake.eclass
```diff
@@ -412,6 +409,7 @@ cmake_src_prepare() {
 			chmod -R a-w "${S}"
 		fi
 		_CMAKE_PREPARE_HAS_RUN=1
+		eqawarn "QA Notice: EAPI=7 ebuild will throw unsupported EAPI error after 2025-11-01."
```
thesamesam, line 412: OK. Once ::gentoo is free of them, consider upgrading to a global-scope warning for the benefit of overlays, like in 47d7fb1a65af001085f973704dfeed320111ab6b.

PR #1095, opened 2026-04-17
Enable python3.14

parona-source:
Ran test+build for the latest version of each.


parona-source:
Looks like im rebasing it a second time today :)

PR #1098, opened 2026-05-20
cmake.eclass: Unconditionally set CMAKE_SYSTEM_NAME in toolchain file

a17r:
Also explicitly add CMAKE_CROSSCOMPILING=(TRUE|FALSE).

Bug: https://bugs.gentoo.org/975603

### DIFF
--- a/eclass/cmake.eclass
+++ b/eclass/cmake.eclass
@@ -358,17 +358,15 @@ cmake_src_configure() {
 	local -x ASMFLAGS=${CFLAGS}
 	local -x PKG_CONFIG=$(tc-getPKG_CONFIG)
 
-	if tc-is-cross-compiler; then
-		local sysname
-		case "${KERNEL:-linux}" in
-			Cygwin) sysname="CYGWIN_NT-5.1" ;;
-			HPUX) sysname="HP-UX" ;;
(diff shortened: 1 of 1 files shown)


eclass/cmake.eclass
```diff
@@ -358,16 +358,14 @@ cmake_src_configure() {
...
-		case "${KERNEL:-linux}" in
-			Cygwin) sysname="CYGWIN_NT-5.1" ;;
-			HPUX) sysname="HP-UX" ;;
-			linux) sysname="Linux" ;;
-			Winnt) sysname="Windows" ;;
-			*) sysname="${KERNEL}" ;;
-		esac
-	fi
+	local sysname
+	case "${KERNEL:-linux}" in
```
thesamesam, line 362: Oh, I forgot to cherry-pick the Hurd change to ::kde, apparently.

thesamesam (approved):
I think this should be okay. The only concerns I would have are related to cross-compiling where I think `KERNEL` is quite a crude mechanism, and any such issues are pre-existing and not made worse by this. Thanks.

I do think it's kind of a feature that it detects broken personality though ;)

floppym (approved):

PR #1099, opened 2026-05-24
net-p2p/ktorrent: metadata: use global USE=upnp description

fkobi:
https://packages.gentoo.org/useflags/upnp

That will do just fine


a17r:
missing sign-off, so I'll commit myself. thanks!

PR #1101, opened 2026-08-09
kde-frameworks/extra-cmake-modules: Allow ECMQueryQt lookups to be cached //and make use of it from ecm.eclass

a17r:
~~TODO: Decide what to do with BDEPEND on dev-qt/qtbase.~~


kde-frameworks/extra-cmake-modules/files/extra-cmake-modules-6.29.0-dont-needlessly-query-qt.patch
```diff
@@ -0,0 +1,94 @@
...
+Date: Sun, 9 Aug 2026 18:36:08 +0200
+Subject: [PATCH] ECMQueryQt.cmake: Do not needlessly query Qt for variables if
+ "_CACHED"
+
+If a ${qt_variable}_CACHED has been set, we will assume this to be the correct
+Qt installation path as defined by i.e. packaging, making a call to qmake or
+qtpaths unnecessary.
+
+ECM may be used by small packages that otherwise do not (or just optionally)
+need Qt for tooling nor build.
```
eli-schwartz, line 12: Should we also specifically call out the fact that this makes cross compilation easier? At least for upstreaming the patch that may help make it an easier sell.

eclass/ecm.eclass
```diff
@@ -47,21 +47,45 @@ if [[ ${EAPI} == 8 ]]; then
...
+			ECM_AUTODEPS=base ;;
+		false)
+			ECM_AUTODEPS=gui ;;
+		*)
+			eerror "Unknown value for \${ECM_NONGUI}"
+			die "Value ${ECM_NONGUI} is not supported"
+			;;
+	esac
+	fi
+fi
```
eli-schwartz, line 68: This is badly indented and it took me a couple tries to realize this was also the closing fi for EAPI==8.

kde-frameworks/extra-cmake-modules/files/extra-cmake-modules-6.29.0-dont-needlessly-query-qt.patch
a17r: I've mentioned it upstream, we'll see.

eclass/ecm.eclass
```diff
@@ -650,8 +679,13 @@ ecm_src_configure() {
 		set(ECM_DISABLE_GIT ON CACHE BOOL "") # *-disable-git-commit-hooks.patch
 
 		set(BUILD_WITH_QT6 ON CACHE BOOL "") # QtVersionOption.cmake: Hard-require Qt6
-		# KDEInstallDirs6 section
-		set(KDE_INSTALL_USE_QT_SYS_PATHS ON CACHE BOOL "") # install mkspecs in same dir as Qt stuff
+		set(QT_HOST_LIBEXECS_CACHED ${BROOT}$(qt_get_libexecdir 6) CACHE PATH "") # FindQtWaylandScanner.cmake
```
thesamesam, line 682: Huh, LIBEXECS _isn't_ a typo.

